PR #284, opened 2015-01-02
BULK - REPT PND1 list.vbs

ramonalisa1:


theVKC:
I'm grabbing these from your repository, but will likely make some adjustments, then pull back to you. Tentatively, it looked good, but I'll check out your comments and see what we can do. Thanks, Lisa!

PR #313, opened 2015-01-08
removed additional tranfers

CDPotter:
I've added a new EDRS check script. 

It works by entering a case number and member number (can find case numbers and can accept 1 digit memb numbers)

Then it grabs SSN and name running it through EDRS and message boxes the worker the results. Pretty basic at this point while we await processing instructions regarding case notes, what to count. Plenty of enhancement ideas to be had.


theVKC:
@CDPotter, I'm going to close this pull request without accepting it, because it contains two changes for two separate issues (eDRS and LTC asset transfer), but coming from the same branch. This causes an issue if I ever need to regress a change (it would change both back, causing lots of issues potentially).

If you could, please:

1. Make a copy your changed files.
2. Revert your repository back to where it was before you committed these scripts.
3. Create two branches, one for each issue.
4. Paste your copy of the changed files into those branches, and then pull request from that branch into master.

The usual process is outlined in the new "How to use GitHub like a rock star" document, which I just uploaded to SIR yesterday. Thanks Charles!

PR #314, opened 2015-01-09
NOTE - MFIP Sanction - DWP Disqualification

GayS:
for review


theVKC:
This looks really good! You can drop it into your agency customized folder if you want to test using it, if you haven't already.

I'll bring it to my mfip contact to see if anything else should be incorporated, but from my knowledge it looks really good. I'll pull it in when I get back to a desk (on mobile now).

GayS:
Great – thanks.  I just posted on FB that github was more difficult than writing the script!  ☺

Let me know if you have changes.

Have a good weekend.
Gay

theVKC:
You get used to GitHub, believe me. So much better than the alternative, which would be emailing constantly... :)

theVKC:
Sent you an email today about some feedback from my MFIP liaison here. Please let me know if you'd like to make the changes, or if you'd like me to pull this in and then I'll make the changes. They were minor and simple (in case you find this comment before you read my email). :)

GayS:
Hi –
Sorry yesterday was a bit on the “crazy” side. ☺   I will look at the suggestions and see what I can do and let you know.

Thanks
Gay

theVKC:
This looks really, really good. Now, the only things I can think of are extremely minor:

1. Can it be both an ES and CS sanction? If it can only be one or the other, this should be a dropdown, not checkboxes.
2. I'm guessing the sanction percent can only be 10%, 30%, or 100%, and can't be all three. Assuming that's correct, this should be a dropdown, not checkboxes.
3. The "FIAT" checkbox could use a bit more description, both on the dialog and in the case note. I'd say something on the dialog like "Check here if you FIATed this case", and in the case note I'd say "Case has been FIATed."

Other than those small issues, this is a GREAT script. So, the question is, do you want to make those changes? Or would you like me to?

GayS:
Thanks, in answer to your questions:

1.      I changed the original checkboxes for ES/CS and the percentage to textbox so that it would also fill in the title.  As long as both are entered into the type of sanction field, both with display in title.

2.      Yes the 10; 30 & 100% are the only choices – I can change that to a drop-down to limit response vs. textbox.

3.      I will expand the “FIAT” language.
I have some time on Friday to make these updates.

theVKC:
Woohoo! Awesome! Looking forward to it! Scripting is fun!! :)


Caution: This e-mail and attached documents, if any, may contain information that is protected by state or federal law. E-mail containing private or protected information should not be sent over a public (nonsecure) Internet unless it is encrypted pursuant to DHS standards. This e-mail should be forwarded only on a strictly need-to-know basis. If you are not the intended recipient, please: (1) notify the sender immediately, (2) do not forward the message, (3) do not print the message and (4) erase the message from your system.

GayS:
I think this one is “done” for now and I can move onto the next scripting project!!  Yeah!

PR #322, opened 2015-01-14
Edrs branch

CDPotter:
Adds EDRS check script

Allows workers to enter a case number and a memb number. Then it will give the worker a message box after seeing if the memb has an entry in EDRS based on their SSN and then their name.


theVKC:
Hmmmm. Looks like this commit had the transfer dialog in there. Perhaps you missed a revert? Can you revert in your branch and try one more time?

CDPotter:
Ok think I found them all and reverted them.

theVKC:
This looks great! Thank you!

PR #323, opened 2015-01-14
Ltc transfer branch

CDPotter:
fixed tranfer to be transfer


theVKC:
Shoot. This one deletes check EDRS. Can you merge from the one branch to LTC transfer branch? This happened because of the revert, and should be easy to fix this way...

CDPotter:
Ok, sorry I thought you wanted two branches based off of the main reverted one, one with the LTC fix and one with the EDRS and no overlap. Makes more sense that you would want a sequential one. I've added a revert to the EDRS revert so now this branch should have both the EDRS add and tranfer fix. If I'm still doing this wrong I can try again and make a new branch!

theVKC:
That's better. Here's what happened: I think you reverted IN this branch, and that caused it to send the revert in the pull request.

CDPotter:
Yup in my haste last night that is exactly what I did. I think I get how it works much better now so hopefully things will be smooth going forward. Sorry for the headache!

PR #332, opened 2015-01-15
Updated County Custom Scripts File

lpshanley:
Updated script file to handle up to 75 custom scripts with a dynamic dialog. This is up from a previous 25. After about 25 previously the dialog would extend beyond a reasonable window size and become unusable. This now supports 3 columns of 25 scripts. This can be expanded later if need be. This may be later updated to facilitate a dynamic multiplier value rather than the current static values. At this time there is no need to larger than a 75 file counter. 

Next update should be able to handle foldering as well. This would in theory create an 'infinite' custom space.


theVKC:
Looks great! Tested it with 50 scripts, no issues. And that'd be an extreme overkill for a county, I'm betting. Nice work!!!

lpshanley:
Oh my goodness yes lol any county with more than 50 scripts of their own should either be using these or should hop on board with the contribution train!

PR #376, opened 2015-02-05
Burial asset error/bug

CDPotter:
Outdated code was causing data to overwrite itself. Updated code a bit.

### DIFF
--- a/Script Files/NOTES/NOTES - BURIAL ASSETS.vbs
+++ b/Script Files/NOTES/NOTES - BURIAL ASSETS.vbs
@@ -518,13 +518,12 @@ IF insurance_policy_number <> "none" THEN
 	call write_editbox_in_case_note("Info on BFE", insurance_BFE_steps_info, 3)
 END IF
 IF type_of_burial_agreement <> "None" THEN
-	If applied_BFE_check = 1 then EMSendKey "* Applied $1500 of burial services to BFE." & "<newline>"
-	EMSendKey "* Type: " & type_of_burial_agreement & ". Purchase date: " & purchase_date & "." & "<newline>"
-	EMSendKey "* Issuer: " & issuer_name & ". Policy #: " & policy_number & "." & "<newline>"
-	EMSendKey "* Face value: " & face_value & "<newline>"
-	EMSendKey "* Funeral home: " & funeral_home & "<newline>"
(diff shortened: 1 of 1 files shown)


Script Files/NOTES/NOTES - BURIAL ASSETS.vbs
```diff
@@ -518,13 +518,12 @@ IF insurance_policy_number <> "none" THEN
...
-	EMSendKey "* Type: " & type_of_burial_agreement & ". Purchase date: " & purchase_date & "." & "<newline>"
-	EMSendKey "* Issuer: " & issuer_name & ". Policy #: " & policy_number & "." & "<newline>"
-	EMSendKey "* Face value: " & face_value & "<newline>"
-	EMSendKey "* Funeral home: " & funeral_home & "<newline>"
-	EMSendKey "--------------SERVICE--------------------AMOUNT----------STATUS--------------" & "<newline>"
-	new_page_check
+	If applied_BFE_check = 1 then CALL write_variable_in_case_note("* Applied $1500 of burial services to BFE.")
+	CALL write_variable_in_case_note("* Type: " & type_of_burial_agreement & ". Purchase date: " & purchase_date & ".")
+	CALL write_variable_in_case_note("* Issuer: " & issuer_name & ". Policy #: " & policy_number & ".")
+	CALL write_editbox_in_case_note("Face value", face_value, 3)
```
theVKC: This could be somewhat problematic, as write_editbox no longer does the variable for indenting. It now always defaults to 6... Can you test that and let me know if it's acceptable?

CDPotter: Sure, we are going to just change them to write bullet and variable function instead. Better to update things while we are in here.

theVKC: Good deal!

CDPotter:
Ok synced with the newer dialogs added.

theVKC:
I can pull it, but let me just make sure: you're OK with an indent of 6 instead of 3?

theVKC:
Also: can you report a bug in the issue list, so that this can be tied as a solution?

CDPotter:
That should be fine. it will bring the script more in line with others.

theVKC:
Resolves #378.

Issue #368, opened 2015-01-30
CAF (NOTES #-G) footer month default to month viewing


ALJ10:

ALJ10:
I accidentally submitted this without an explanation. Oops! Anyways, workers in Crow Wing County are wondering if it's possible to have the CAF script default to the footer month they are viewing instead of having it default to the next month.

ALJ10:
Afterthought: Either have it set to the month they are viewing or the current footer month. Hope that makes sense!

theVKC:
It does. And this is actually the design, it's a known bug that it doesn't do it. However, even though it's known, I never added it to GitHub for some reason! Anyway, now it's there. :)

lpshanley:
Im going to start working on this one!

lpshanley:
Done and commited

theVKC:
I think I see what's going on here... you didn't do this on a separate branch. When you work on two separate script ideas (like AGENCY CUSTOMIZED and CAF), you need to make two branches from master, one for each change. That way, if something ever goes wrong, I can revert the one without affecting the other. Does that make sense?

Here's how I think you solve this dillema:

1. Copy your changed code for both scripts. Save it somewhere you'll remember it.
2. Revert all commits. If that doesn't work, delete your fork entirely.
3. Pull request MN-Script-Team's main repo back into yours.
4. Create two branches, one for AGENCY-CUSTOMIZED and another for CAF.
5. Copy each script change you made (and copied in step one) back into its respective branch.
6. Pull request each branch.
7. When I accept each branch, go ahead and delete them, but only after I accept them.

I know it might seem goofy, but I'd never want to inadvertently introduce a bug by reverting some CAF commit in six months which somehow destroys AGENCY CUSTOMIZED. Does that make sense? It's about preserving the commits that relate to each other inside branches which can be easily reverted if anything goes goofy.

Otherwise, your changes look solid. And, they are VERY appreciated.

lpshanley:
Now its done AND in its own repo!!!

ALJ10:
Now that this has been fixed, do I need to run the config tool again? We are a beta county so I was thinking our scripts would automatically update. But when I went to use the CAF (just to test), it still shows next month as the footer month.

Let me know when you get a chance. Pls & ty!

theVKC:
[...]

PR #383, opened 2015-02-06
I reverted all the old changes. This should just be the CAF script.

lpshanley:


lpshanley:
just look at the files changes thingy
Not sure how to ditch the extra commits

lpshanley:
Added folder support

Script Files/NOTES/NOTES - CAF.vbs
```diff
@@ -29,11 +29,9 @@ ELSE														'Error message, tells user to try to reach github.com, otherwi
 END IF
 
 'DATE CALCULATIONS----------------------------------------------------------------------------------------------------
-next_month = dateadd("m", + 1, date)
```
theVKC: This looks like a change that didn't get reverted. Not sure what to make of it... any insight?

theVKC: Never mind, I explained what you need to do on issue #368. Hope that helps. If it doesn't, let me know. Or, possibly, let @CDPotter know, because he dealt with a similar issue a few weeks ago (and handled it like a pro).

PR #417, opened 2015-02-24
County custom script revision

lpshanley:
Added code to make the arrays more functional. At this time it appears this code is solid. The only issue we have been experiencing is while using GitRaw. I would like to see if this can be resolved by merging this over the current version.


lpshanley:
Just added the last out of range i fix :-)

PR #438, opened 2015-03-02
BULK - MISC NON-MAGI HC DEDUCTIONS

ramonalisa1:
This script will look at the PDED panel for the lesser known hc deductions, such as pickle, disabled child, etc, and put the codes into a spreadsheet so coding errors can be detected.


theVKC:
Sorry @ramonalisa1, I ran into an issue with this one... it's not displaying the dialog. It looks like you've got `Dialog SWKR_list_generator_dialog` when you should have `Dialog misc_non_magi_hcdeduction_list_generator_dialog` in line 63. Any chance you can fix that and resubmit? If you have questions, shoot me an email. I should be relatively available the next few work days. :)

Issue #447, opened 2015-03-03
NOTES - DOCS RECEIVED no longer puts the type of document in the header


RobertFewins-Kalb:
Instead, it's pushing the document received to the second line of the case note. I have received some complaints about this. If it's not a data privacy issue, I would like to commit a change to move it back to the header

theVKC:
How/when did that happen? In any case, I can't imagine it's a problem to set it back the way it was. I might do a beta push today for a few new features to test, so by all means proceed. While you're in there, can you update the case noting functions and copy the new FuncLib URL logic from the CAF script in there? We need to make those changes incrementally as we work...

courtrightd:
Any update on merging this into BETA?

theVKC:
We have a fix but it isn’t enough of a priority for me to push it out before the rest of the beta stuff. Planning a beta release in a week or so.


Caution: This e-mail and attached documents, if any, may contain information that is protected by state or federal law. E-mail containing private or protected information should not be sent over a public (nonsecure) Internet unless it is encrypted pursuant to DHS standards. This e-mail should be forwarded only on a strictly need-to-know basis. If you are not the intended recipient, please: (1) notify the sender immediately, (2) do not forward the message, (3) do not print the message and (4) erase the message from your system.

courtrightd:
Thanks!

PR #451, opened 2015-03-03
Update to DOCS RECEIVED

RobertFewins-Kalb:
(1) Updated the header to put the document received in the header
(2) Updated the url to allow for the BETA FuncLib
(3) Added new functions to the case noting


RobertFewins-Kalb:
NOW, this should resolve #447

PR #457, opened 2015-03-03
DWP budget

LaLarson:
This is a case note that will be used prior to approval when someone is referred to ES after turning in all verifications, and determining that eligibility exists.


theVKC:
This looks AMAZING! Resolves #458.

PR #474, opened 2015-03-18
Create NOTES - BABY BORN.vbs

LaLarson:
This is a case note template for MAXIS, when someone informs the agency that they had a baby/want to add a new baby to the household.


theVKC:
Please replace "select one"s with looping to require that they not say "select one". If that makes sense at all, IDK, my caffeine is running low.

PR #475, opened 2015-03-18
NOTES - CHANGE REPORT FORM RECEIVED.vbs

khiestand:
This is a script that Financial Workers can use to enter information they have received on a Change Report Form DHS 2402.  The script will case note info entered and also has the option to navigate to DAIL/WRIT to create a new blank TIKL message, automatically sets the TIKL date out 10 days from the date the note is created.


theVKC:
I'm going to close this for the following minor reasons, if you could resubmit that'd be swell:

* Can you set the "select one..." to require one of the options to be selected?
* Can you either offer an editbox to create TIKL info, or explain that it's heading to a blank TIKL?

PR #478, opened 2015-03-18
NOTES - GAS CARD ISSUED.vbs

khiestand:
This is a script counties can use if they issue gas cards to clients.  It could also be modified to use for bus vouchers, etc.  This is one that Wright County uses, and our increments are in 10, 20, 30 or 40 dollar amounts.  Clients can only receive so much in a given calendar year, and they can only receive $40 every 30 days.  The note is nice as workers can give cards out to clients not on their own caseloads in a backup function and use this script so every note is standardized.  That dropdown box could be changed to be an edit box instead.  Just an idea, Wright County will use in their Agency Customized folder.


theVKC:
Same deal, can you make the select one mandatory?

PR #480, opened 2015-03-18
NOTES - Medical Opinion Form Received

CDPotter:
a MOF case note.


theVKC:
There are a couple of very minor issues with this one... so minor I'm going to gloss over them, @CDPotter. I want this script really badly to go live to the beta group tomorrow (it's been on the list for for-ev-er), so I'll make the changes myself. But, here they are summarized, just for your info:

1. You don't need to add the colon in `write_bullet_and_variable`, it'll do that for you.
2. Your first editbox was inadvertently named the default instead of case_number.

I'll fix those in a separate commit. Otherwise, job well done. Perhaps future enhancement will have it update DISA and WREG (for SNAP folks). #EnhancementRequest

PR #481, opened 2015-03-18
fraud info

CDPotter:
fraud info case note.


theVKC:
I like it! FYI, the colon thing on the other request applies here, @CDPotter. I'll go ahead and add it. Nice work!

CDPotter:
A good example of work smart not fast. I totally missed that it was doubling up on the colons. Thanks for helping to fix that and letting me know!

PR #482, opened 2015-03-18
NOTES - DRUG FELON.vbs

devonne01:
This script would be utilitized for clients that are on SNAP who have been convicted of a drug felony in the last 10 years and need to comply with UA's, this script would be able to track their UA's and if there are any changes to their benefits in results of the UA tests.


theVKC:
Simple case note template. I'll take a look!

theVKC:
I'm going to close this since the checkboxes aren't working. If you can get the checkboxes working, it's in. :)

PR #488, opened 2015-03-18
MEMOS - OVERDUE BABY

LaLarson:
Creates a SPEC/MEMO to send when someone has not reported the birth of their baby/end date. Also creates a CASE/NOTE, and optional TIKL.

### DIFF
--- a/Script Files/MEMOS/MEMOS - OVERDUE BABY
+++ b/Script Files/MEMOS/MEMOS - OVERDUE BABY
@@ -1,0 +1,129 @@
+'STATS GATHERING----------------------------------------------------------------------------------------------------
+name_of_script = "MEMOS - NOMI.vbs"
+start_time = timer
+
+
+'LOADING ROUTINE FUNCTIONS FROM GITHUB REPOSITORY---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


theVKC:
Wow, @LaLarson! This is amazing. Here are my very minor notes:

1. Name of script was set as "NOMI" when it should be "OVERDUE BABY". I can change that.
2. The URL for the routine functions is set to master when it should be BETA. You would have had absolutely no way to know that was a thing. Basically, there'll be new code to copy/paste after tonight, I hope. I'll change it for now, no worries.
3. Your case note function puts everything into CASE/NOTE but doesn't indent additional lines, or include that cool bullet. I'm going to modify it to include that, just to stave off any accessibility issues. Very minor issue, and you can see the changes on lines 101-103 of the script when I'm through.
4. The TIKL at the end should send it (add a transmit and a PF3 in the end), and include a "success" message to let the worker know everything it did (I'll add that myself). We never covered that, frankly you're getting loads of extra credit for making a memo instead of a case note during hackathon. #LauraIsAHero

I'll go ahead and make these changes for you, no worries. I'm eager to get this out there. Laura_is_amazing = very_yes

PR #490, opened 2015-03-18
Patch 1

LaLarson:
Added "yes" "no" logic.


theVKC:
Nailed. It. Only thing I'd change is the master FuncLib reference to be BETA (URL on line 6), and you wouldn't have known I wanted you to do that. Nice work, @LaLarson!

PR #494, opened 2015-03-19
Create NOTES - LTC TRANSFER PENALTY

IlseFerris:
LTC Transfer Penalty case note.  Gives template to LTC worker in which to case note the details of a transfer penalty.  Script will also case note and automatically case note the lookback period based on the baseline date.  Case note also pulls in the Uncompensated Transfer Calculator from MAPT, and case notes this information.


theVKC:
Love it! Looks great!

PR #516, opened 2015-03-25
BULK - NON-MAGI HC DEDUCTION

ramonalisa1:
This is working well now.  Thanks @RobertFewins-Kalb for the help.  I really appreciated it!

### DIFF
--- a/BULK - MISC NON-MAGI HC DEDUCTIONS
+++ b/BULK - MISC NON-MAGI HC DEDUCTIONS
@@ -1,0 +1,256 @@
+'GATHERING STATS----------------------------------------------------------------------------------------------------
+name_of_script = "BULK - MISC NON-MAGI HC DEDUCTION.vbs"
+start_time = timer
+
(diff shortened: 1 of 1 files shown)


ramonalisa1:
Just out of curiousity .. is there a way to get the spreadsheet to format a column as dollars/cents autormatically or to autoformant column width?

Also, when I run this from MAXIS Script Play it works fine, but if I use the Agency Custom button, it says the script is done right away but nothing is on the spreadsheet.  Thoughts?

RobertFewins-Kalb:
(1) To format in dollars, you can use the command FormatCurrency.

(2) RE: column width, you could try to GTS, but I think the way you have it set up is fine.

(3) RE: Agency Custom...I adjusted the location to where the script is depositing the worker_number. It was dumping it at the start of the line (in REPT/ACTV, for example) and I had it dump just the last three characters. You may need to move the location where worker_number is being dumped.

PR #520, opened 2015-03-27
Ltc transfer penalty

IlseFerris:
Updated baseline date logic to error if worker does not enter a date in MM/DD/YYYY date format.


theVKC:
Pull request made in error, no files changed, removing.

PR #551, opened 2015-04-09
Updated FuncLib block on all redirects and script files

### LINKED ISSUE #468: Announcing Project Cerberus: RELEASE/BETA/master branches for FuncLib
It only makes sense, especially in light of this week's FuncLib navigate_to_MAXIS_screen debacle. This would involve amending the FuncLib call to include all three branches as follows:

* Scriptwriters would use master (i.e. `beta_agency = ""`)
* Beta agencies would use BETA (i.e. `beta_agency = True`)
* Everyone else would use RELEASE(i.e. `beta_agency = False`)

For now this will be a single issue, however, if it generates bugs we will create other issues in this milestone too. Anyone feel like taking this one on?

theVKC:
Also includes the new run_locally option. This will require new instructions to be issued to scriptwriters. Resolves #468.


IlseFerris:
WHOA!

PR #579, opened 2015-04-22
NOTES - CLIENT CONTACT

GayS:
Merged client contact and client contact (call center) scripts into one script with a secondary dialog box for the call center specific info.   Added the following updates:  auto-fill date & time, options explicit, case_number_finder, navigate_to_MAXIS_screen, and write_variable/write_bullet functions.


theVKC:
I think we have a logic issue in this script. Upon testing it, I've run into a problem with contact_type. Did you merge both dialogs together? It looks more like a checkbox for call center work, and not a merging of the two scripts together per se. But, perhaps I'm missing something. In any case, feel free to re-evaluate and resubmit if you'd like.

Issue #562, opened 2015-04-13
BNDX Scrubber


RobertFewins-Kalb:
Enhancement Request: Clean-up the logic of the BNDX Scrubber so it isn't a gigantic mess. Seriously, @RobertFewins-Kalb , what were you thinking?

Use arrays to compare multiple RSDI claims instead of that messing smushing.

RobertFewins-Kalb:
I'm on it, @RobertFewins-Kalb . Please do not be so condescending.

theVKC:
Settle down you two.

RobertFewins-Kalb:
Resolved with Pull Request #580

PR #581, opened 2015-04-23
Bndx scrubber update

RobertFewins-Kalb:
The actual factual update.


RobertFewins-Kalb:
should ACTUALLY resolve issue #562

PR #598, opened 2015-04-30
Revw/mont patch 1

CDPotter:
added Ucase logic to make sure first page isn't skipped for workers with letters in their x#s


theVKC:
Hey @CDPotter, was this an existing issue on the list?

CDPotter:
I added the issue after discovering and fixing it.

PR #610, opened 2015-05-21
Caf pw  check patch

CDPotter:
Added 2nd PW check after dialog. This will allow workers to not worry about getting passworded out while writing a detailed case note.


theVKC:
Resolves #611. Nice work.

PR #659, opened 2015-06-02
Project louis part 1

RobertFewins-Kalb:
BIG update...

The script builds Household_array. The map of the array starts on line 152.

The script is build to only look at panels for household members that are relevant. For example, when looking at UNEA, the script only looks at UNEA for those household members that are children on PARE 01.

The script is programmed with a hard break on ELIG/DWP when FIATing to make sure that it does not jump ahead of MAXIS and dump information that is not relevant to the case.

MFIP is a much simpler ordeal.

Possible enhancement opportunity is multiple benefit month handling. That requires restructuring the script to pull the active or pending cash program off STAT/PROG to enable the script to handle the transition from DWP to MFIP. Currently, the script looks at CASE/CURR to determine the CASH program, but that is inadequate for multiple benefit month handling.

Questions?


theVKC:
Looks really good.

PR #682, opened 2015-06-09
CSES-FIX

lpshanley:
Corrected an issue on CSES that was resulting in the script stopping. The script was converting any value with an amount in the thousands to use a comma. This was causing the script to error as UNEA does not allow this as a value. Changed the format of FormatNumber(value,2) to FormatNumber(value,2,,,0) so that the script should now disallow the use of commas but should retain the 2 decimal places. (EX 1,000.98 = 1000.98).


theVKC:
I know this is unusual, but I'm actually going to deny this and pull it instead to a working branch I have set up for migrating away from Excel. Nice catch!

theVKC:
(You'll still get the credit that way instead of me getting the credit for this mod)

theVKC:
OK, never mind because I'm getting a pull error. I'll do it the old fashioned way.

lpshanley:
Problem Solved, Crisis Averted ☺

PR #683, opened 2015-06-09
Create BULK - ES DB Maintenance.vbs

courtrightd:
This is a very early version of a script to populate the ES database from rept/mfcm.


courtrightd:
Current versions of ES Referral, status update and the database maintenance scripts that function in SLC to read and write to DB.  Also a template of the access database.  The Status Update script includes dialogs that pull data from database.

PR #695, opened 2015-06-11
Privileged case handling added

### LINKED ISSUE #646: BULK - REVW/MONT closures can't handle PRIV cases in banks
When the script is run on a case bank, it errors out and stops when it hits a priv case that the worker doesn't have access to.  Major problem when it is case number 156 of 640. I'm thinking add logic to read screen on each case for the PRIV error message, if yes then dump case number to a file and skip to the next one.

courtrightd:
Resolves #646.  The script will now bypass privileged cases and give the worker a list of the ones it skipped in a msgbox at the end.


theVKC:
This looks FANTASTIC. Is it tested? I presume yes, but I would not be doing my job if I didn't ask. :)

courtrightd:
Yes, it is tested and working, though I only found 1 PRIV scenario for each report to test, so don’t get mad at me if next month a new issue is found.  ☺

theVKC:
OK, well I'm going to work it in then. :)

PR #697, opened 2015-06-11
Bulk find maepd medi cei

RobertFewins-Kalb:
This script looks through your REPT/ACTV to create a list of clients that are active on MA-EPD and eligible for Medicare Part B Reimbursement through the county.


theVKC:
Looks great! Remind me which issue(s) this solves?

Issue #684, opened 2015-06-09
LTC intake approval script issue with 'application date'


IlseFerris:
Looks like the edit box is not descriptive enough.  Needs to refer users to update the baseline date instead of the application date.  Also has some updating to do.

Oh, and I got this.

theVKC:
Good luck @IlseFerris!

theVKC:
Marking as critical because there was some functional problems with it...

PR #702, opened 2015-06-12
Ltc initial approval fix

IlseFerris:
* added proceed_confirmation
* updated case note header, & to case note if no * spenddown exists
* updated old custom functions
* updated dialog to have more specific MsbBox's    for errors.  
* Updated case note to note both the lookback period start and end date. 
* Updated MsgBox if date is not MM/DD/YYYY format.
* Added in other custom functions (cancel_confirmation)


theVKC:
Resolves #684. Going to also pull this into BETA for testing for a mid-month critical fix release.

Issue #401, opened 2015-02-18
Asset Assessment Note Script


tdahl21:
Hey All,

There seems to be an issue with the second page of the Marital Asset portion of the asset assessment script.  A coworker ran into an issue where the second page of the marital asset portion was not being written into the person note nor the Maxis casenote.   Thank you for checking into it!

theVKC:
Hmmm. I don't know of any issues with it. Of course, we have a case note function issue right now, which is being tested with Anoka and Goodhue (other beta counties should get the fix in the next day or two).

Any chance I can ask you to ask your coworker to try it again, and if it happens again, email me a case number to look at? I can dig into it and see what we can uncover, possibly tomorrow.

tdahl21:
Hey Veronica,

I'll send you the case # through SIR.

IlseFerris:
I'm pretty sure that the AA case note does not copy more than one page.  I can take this one

IlseFerris:
I can work on this tomorrow if that works for you

theVKC:
Sounds great! :)

PR #709, opened 2015-06-15
Updated to read up to 4 total marital asset screens & note them

IlseFerris:
.  There are up to 8 screens available, and time permitting will be updated with an array in the future.  also updated the case note to NOT case note empty blank asset spots into person or case note, updated old functions with new custom functions.


IlseFerris:
This resolves issue #401

PR #718, opened 2015-06-17
Ltc intake approval fix

RobertFewins-Kalb:
The script was calling the wrong function. This was causing the error reported by @khiestand


theVKC:
Repairs #719.

Issue #515, opened 2015-03-25
LTC/GRH list gen - cutting off last few names on the list


theVKC:
That's the rumor. I have specific worker numbers. @IlseFerris has offered to take this on.

IlseFerris:
I can work on this tomorrow if that works for you

theVKC:
Again, sounds great! :)

theVKC:
@RobertFewins-Kalb actually found a fix for this and is pulling it.

PR #721, opened 2015-06-17
Update BULK - LTC-GRH LIST GENERATOR.vbs

RobertFewins-Kalb:
Moved last page check before the nested do.

When you get to the last page on REPT/ACTV, "THIS IS THE LAST PAGE" is displayed. The way the script was programmed, it was reading "THIS IS THE LAST PAGE" immediately after the PF8 and it prevented the last page from being read.


theVKC:
#515 is resolved.

Issue #722, opened 2015-06-18
Critical: Bulk cola scrubber GRH issue


CDPotter:
So apparently the GRH autoapproved message is INFO type and not COLA. @RoyStearns  or @RobertFewins-Kalb  Would either of you be able to look at it and account for this change since you're both more familiar with the code? (I know Robert is out today). I can email a image of the message or there are some examples in x102524

CDPotter:
Also it seems to be reading the COLA message for the wrong case. This script is very much down.

CDPotter:
Also MSA autoapproved is also a INFO message.

IlseFerris:
Thanks Charles.  I will let users here know.

IlseFerris:
Does anyone have this?  I'd say this is critical

CDPotter:
I'm trying to work through it but it's not going smoothly or quickly.

IlseFerris:
Thanks Charles.  You are my current hero ;D

CDPotter:
Update: I have submitted fixes for these and they will be reviewed tomorrow. Please inform workers to not use them for the time being.

theVKC:
@CDPotter, any update on pull requesting this to BETA and RELEASE? If I don't hear from you soon, I'll just plan on shutting it down.

CDPotter:
I've tried to fork BETA and make the changes but I keep getting unable to merge automatically. I could have sworn I submitted another RELEASE one. I've deleted my current forks for BETA and RELEASE and am planning to submit pulls shortly.

PR #726, opened 2015-06-18
Bulk cola fix

CDPotter:
Won't read same message over and over when bulking
won't set cursor wrong after a maxis check
allows for INFO messages since GRH and MSA are info messages.


CDPotter:
#722

PR #727, opened 2015-06-18
Bulk cola fix

CDPotter:
Won't read same message over and over when bulking
won't set cursor wrong after a maxis check
allows for INFO messages since GRH and MSA are info messages.


CDPotter:
#722

theVKC:
@CDPotter, same problem as the other one.

CDPotter:
Ah sorry, I didn't pay that close of attention when submitting them. Luckily the COLA bulk isn't on RELEASE so no need to worry about that one. I guess only BETA counties have it.

PR #730, opened 2015-06-18
Updated 01/15 to 07/15

CDPotter:
Looks like some formatting for the header was included in the change as well.


theVKC:
Thank you @CDPotter! I will also pull this into BETA. Please don't delete the branch until then.

PR #741, opened 2015-06-19
Fixed multiple issues

CDPotter:


CDPotter:
#722 This method was WAY easier!

theVKC:
It is, isn't it? I don't like it more than command line or for master handling, but it works well for stuff like this.

Issue #792, opened 2015-07-06
LTC burial asset script


CDPotter:
Had a worker here at Anoka suggest that we add some check and balance system to prevent people from accidentally not leaving the burial agreement type as NONE when they don't notice that it is a field. Worker had completed everything except that box and lost a lot of data because he missed the box. 

Thoughts?

IlseFerris:
Thanks Charles!  Do you want to clean up some of the old functions too while you're in there.  Cancel_confirmation is one that stands out from the get go.

CDPotter:
Sure I can close my current pull and fix some of them.

PR #793, opened 2015-07-06
Added safeguard

CDPotter:
Makes sure workers select a type or burial agreement or none.

#792


CDPotter:
Gonna go back and add newer functions.

Issue #753, opened 2015-06-26
Check your notice warning (enhancement)


CDPotter:
Had a suggestion from a supervisor that maybe we could consider adding a reminder at the end of the CAF that would tell people to check their notices make sure they read correctly, and add WCOM if needed. 

My gut says to do it on Approved/denied/closed progs, if we were to add anything. I wanted to pitch it to the crowd and see what other people thought.

courtrightd:
I agree with Approved/denied/closed as the route with this if one is going to add it somewhere.

theVKC:
Agreed. Frankly, if we are to put ANYTHING into the CAF script, I think it should be a warning to use approved/denied/closed programs after running the CAF. #UseTheRightToolForTheJob

CDPotter:
So we're thinking add a reminder message box at the end of the CAF to tell people to use Approve/Closed/Denied progs as necessary, and a message box at the end of those scripts reminding people to check their notices read correctly?

theVKC:
I concur.

IlseFerris:
Agreed.  You got this!

theVKC:
Do you want this, @CDPotter?

CDPotter:
Can do.

PR #797, opened 2015-07-06
Notice reminders

CDPotter:
#753


theVKC:
This looks good, but can you please update the first word of these to read "Success!" with the exclamation point? This is important for when counties do time savings studies.

CDPotter:
Got it!

PR #801, opened 2015-07-07
Added handling for MFIP housing grant

courtrightd:
Also corrects an occasional issue with unapproved versions being read


theVKC:
I like it. Nice job!

Issue #802, opened 2015-07-07
CRITICAL BUG -- Syntax Error with SSA Agreement Handling 


RobertFewins-Kalb:
A financial worker reported that the SEND SVES script is not working. When I looked at the code, the closing message for the script end procedure inserted to get around the SSA Agreement does not have the closing text in quotes, causing a syntax error.

RobertFewins-Kalb:
Resolved with #805

PR #805, opened 2015-07-07
SSA agreement fix

RobertFewins-Kalb:
Added double-quotes around the SSA agreement script end procedure. The bug is creating a syntax error.


RobertFewins-Kalb:
Resolves #802

PR #812, opened 2015-07-08
committing to pull request into BETA

IlseFerris:
critical bud fix for FSET sanction button


theVKC:
So this one totally looks like it works, but it'll be superseded by Robert's change. But, you should do this exact thing, but send it to RELEASE instead.

PR #814, opened 2015-07-08
Patch 2

IlseFerris:


theVKC:
Same problem as before, dear. You have all the files here. If you did this from GitHub for Windows, I recommend not doing that when pushing into RELEASE.

Issue #835, opened 2015-07-13
LTC- MA approval enhancement idea


IlseFerris:
Add checkbox "is this for a paperless IR", and if so, add this verbiage to the case note header.

I can take this one.

theVKC:
Did you do this one already, @IlseFerris?

IlseFerris:
Yep, pull requested it this morning (#843)

PR #843, opened 2015-07-15
Ma approval-added paperless IR option

IlseFerris:
So you didn't like MAXIS_MAXIS?  The system so nice I named it twice?

Also cleaned up script and old functions.


theVKC:
Resolves #835.

Issue #774, opened 2015-07-01
EMA mnsure app processing case note 


IlseFerris:
Hennepin county HC worker has requested this.
"We do a lot of EMA mnsure app processing and would love a case note script for this – probably would just be a spin off of the HCAPP case note."

Sounds pretty legit, any takers?

IlseFerris:
Add the following:
-	Have the heading read something to the effect of ***EMA mnsure app received 6/26/2015: approved***   
-	EMMA panel (put it in the case note near the imig stuff – like right under *Cit/ID:

***EMA mnsure app received 6/26/2015: approved***   
* HH comp: 1a - 48 y/o single undocumented female   
* Cit/ID: Member 01- non-citizen, no verifs provided
* EMMA: Begin Date – 06/26/2015 
> Medical emergency - 01 Phys Or Mntl Sudden Onset
> Health Consequence: 01 Health Jeopardy 
> Verification: OT Other Document  
* Notes on income: no income reported               

Omit the following: 
-	Schl/stin/stec
-	Coex/dcex
-	Assets
-	Stwk

PR #856, opened 2015-07-15
NOTES - LEP - EMA.vbs

devonne01:
This is a case note function that will allow workers to process EMA mnsure applications in Maxis


theVKC:
Resolves #774.

Issue #819, opened 2015-07-09
SNAP E&T letter defaulting to MEMB 01 when 2-digit MEMB number not used


IlseFerris:
I can take this. 

I will also update the case note to include which MEMB name and # is being referred to orientation.  Any other enhancements that anyone can think of? @courtrightd @ruthmaloney @CDPotter

PR #861, opened 2015-07-16
Snap e and t update

IlseFerris:
This function formats a string (y) being entered into a case note. This function retains the functionality from the old function write_new_line_in_case_note(x) but enables for line headers (bullet) and indentation.  The header must be stored in quotation marks. The function automatically places a colon ( : ) at the end of the header. 
The function also checks to determine if the case note needs to go further than 4 pages and will modify the case note header to read (1 or 2) and create a new case note with header and (2 of 2) if necessary.

Resolves issue #819

### DIFF
--- a/Script Files/MEMOS/MEMOS - SNAP E AND T LETTER.vbs
+++ b/Script Files/MEMOS/MEMOS - SNAP E AND T LETTER.vbs
@@ -1,3 +1,5 @@
+worker_county_code = "x127"
+
 'STATS GATHERING----------------------------------------------------------------------------------------------------
 name_of_script = "MEMO - SNAP E AND T LETTER.vbs"
 start_time = timer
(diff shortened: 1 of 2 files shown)


Script Files/MEMOS/MEMOS - SNAP E AND T LETTER.vbs
```diff
@@ -1,3 +1,5 @@
+worker_county_code = "x127"
```
theVKC: This won't work for 86 of the counties, @IlseFerris. :P

theVKC:
Closing this because of `worker_county_code`, @IlseFerris. Please fix that and resubmit. :)

IlseFerris:
Dang it, I knew I'd forget!  On it.

Script Files/MEMOS/MEMOS - SNAP E AND T LETTER.vbs
IlseFerris: Wait, what...

PR #891, opened 2015-07-23
New NOTE script for Proof of Relationship documents received.

kaleykobber:
This script is for case noting proof of relationship documents received for household members.


RobertFewins-Kalb:
@kaleykobber the script needs a script_end_procedure() on the last line for usage statistics

theVKC:
I think we can go ahead and add that this time, but we should include it in the best practices document.

PR #893, opened 2015-07-23
Update ACTIONS - PAYSTUBS RECEIVED.vbs

RobertFewins-Kalb:
Added all dialog variables as arguments through the custom function that loads the paystubs dialog. That should enable smooth sailing.


RobertFewins-Kalb:
If this doesn't work, then we need to stop accepting paystubs

PR #894, opened 2015-07-23
New Script for Lobby No Show Note

C-Love:
Script creates a case note when a client no shows to a lobby interview, includes option to send NOMI. @RobertFewinsKalb


theVKC:
@RobertFewins-Kalb, she meant to send that to you, I think. :)

PR #898, opened 2015-07-23
Create new file

lor03ole:
MNSure Application


lor03ole:
we're having issue with uploading this info onto github.  This is a Curam/HC Application Retro coverage. this is only case note at this time, may not be complete.  Idea was from Ilse.

RobertFewins-Kalb:
I am not going to pull this script into the NOTES main menu for the following reasons:
* I have concerns about object spacing in the dialog
* The TIKL check is not connected to logic to TIKL

PR #906, opened 2015-07-27
Worker number issue

CDPotter:
Worker number was being converted to add worker_county_code to the x_number variable no matter if it was already there or not (for solo runs). Tested on solo and batch and it seemed to work.


theVKC:
Is there a corresponding issue with this, @CDPotter? If so, can you tag it?

CDPotter:
There is now =) #909

Issue #912, opened 2015-07-28
NOTE - Pregnancy Reported


khiestand:
Hi,

I have written a new script this morning called "Pregnancy Reported."  This script will case note the following details:
HH Member Pregnant, Due Date, Conception Date, How the Pregnancy was reported, other comments/notes, Program/s pregnancy reported for, if MMIS was updated, and if verification request for cash was sent. 
The script will also update the stat/preg panel for the member that is pregnant.

Please assign this issue to me, thanks!

theVKC:
Say, I haven't wrapped the BETA release yet! If you want this included, can you pull it in? Or will you need a bit more time? Either way, way to go @khiestand! :)

khiestand:
@theVKC I am working on it with Robert, have a couple of things to clean up on it.  When would you need it by?

RobertFewins-Kalb:
If you intend to have it ready today, please let me know so I can update the NOTES main menu

khiestand:
@RobertFewins-Kalb @theVKC 
I think I have done all that I can do on it.  Might need more experienced scriptwriters to take a peek just to be sure.  If either of you think you will have time today, then great, if not, I can definitely go on the next release.  Thoughts?

RobertFewins-Kalb:
email me

I will update the main menu

PR #913, opened 2015-07-28
NOTES - PREGNANCY REPORTED.vbs

khiestand:
This is a case noting template that case notes details of a pregnancy reported to the agency.  It will also update the STAT/PREG panel for that member.

Having some troubles around line 144, also, when I run this in training region the case number doesn't auto-populate into the dialog.

Issue #912


RobertFewins-Kalb:
@theVKC , if you pull the script, I can submit a finish. @khiestand , I have addressed the issues in your documentation.

theVKC:
Get to it, @RobertFewins-Kalb! And thanks you two!

PR #936, opened 2015-07-30
Update ACTIONS - PAYSTUBS RECEIVED.vbs

RobertFewins-Kalb:
Removed PIC_paystubs_info_adder function and replaced the 1 call with the logic from the function. We may have been too deep down the vbs rabbit hole...


theVKC:
Normally I'd take this, but since there are tons of updates in master at the moment, I'm going to reject it and simply pull all at once.

Issue #943, opened 2015-07-30
Enhancment BULK ADDR report


CDPotter:
It was suggested that we add a field for mailing addresses as well. I'm on it.

theVKC:
This is complete, but I'm waiting until after Monday's script release to send this through to beta/master.

PR #944, opened 2015-07-30
upgrades

CDPotter:
Added mailing address fields, and better comments. #943


theVKC:
@CDPotter, because this is an enhancement and we're testing the existing beta until Monday, I'm going to hold off on this, and will push it out with next week's beta release.

PR #969, opened 2015-08-06
Update MEMOS - SIGNIFICANT CHANGE.vbs

MelissaFox-Stearns:
Thanks for your help guys! @theVKC @RobertFewins-Kalb @IlseFerris
Significant Change script includes a spec/memo denial letter that is automatically sent if you are denying the request and can put out a Tikl if client is requesting for a future month. 
The script is written with the new Do..Loop process which brings up an error message if a mandatory field is not entered.

### DIFF
--- a/Script Files/MEMOS/MEMOS - SIGNIFICANT CHANGE.vbs
+++ b/Script Files/MEMOS/MEMOS - SIGNIFICANT CHANGE.vbs
@@ -46,79 +46,75 @@ IF IsEmpty(FuncLib_URL) = TRUE THEN	'Shouldn't load FuncLib if it already loaded
 END IF
 'END FUNCTIONS LIBRARY BLOCK================================================================================================
 
-BeginDialog SigChange_Dialog, 0, 0, 291, 245, "Significant Change"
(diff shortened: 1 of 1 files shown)


IlseFerris:
Cool script Mel!  You appear to be missing a couple steps after navigating to SPEC/MEMO.  You will need to: 
1. PF5 to create a new MEMO
2. EMWriteScreen "x", (row), (col) ---to select whom to send the memo to
3. Transmit

Then you'll be in edit mode :)

theVKC:
Hey @IlseFerris, from my view her script does this. You might not see it if you're looking at the diff summary instead of the raw file. I always check out the raw file before it goes up, and I see the PF5 et. al.

@MelissaFox-Stearns, it's in and it will be included in the beta. Nice job!

IlseFerris:
I was looking at the summary. Sorry for the scare @MelissaFox-Stearns.  

ANNNDD, I think I'm done thinking for today...ugh....

MelissaFox-Stearns:
LOL @IlseFerris  I know the feeling :)

Issue #868, opened 2015-07-17
BULK - MA-EPD/MEDI list errors in navigating to MMIS.


theVKC:
navigating to MMIS... error prone because it's looking for C402, et. al, and not using the better EMSearch functionality that I believe I put into FuncLib some time ago. @RobertFewins-Kalb, can you look into this for me? If you can't find my functionality in FuncLib or possibly in my "MAXIS to MMIS" script, try looking for RECIPIENT FILE APPLICATION and using that as your function.

RobertFewins-Kalb:
I think I just need to update the logic in the custom function to navigate to MMIS...

Instead of writing C302 or C402, it should be writing ("C3" & worker_county_code) or ("C4"  & worker_county_code) but I'm not entirely sure how to test before pulling a change...

Y-Lor-Y:
If you can give me access to the script, I can test it for you.  I have MA-EPD cases that I can run this through to see if it's functioning.

RobertFewins-Kalb:
I'll submit a fix today. The script is in GitHub, in with the BULK scripts...

Y-Lor-Y:
Ah, found it.  I was looking for BULK - MA-EPD, but it's under BULK - FINDING MA-EPD.

RobertFewins-Kalb:
As it happens, there's an issue with the MAXIS to MMIS and MMIS to MAXIS navigation functions...

I'm thinking about having the script just run in one window. Thoughts...?

Y-Lor-Y:
I'm fine with that.  I'm not sure about other counties, but here at RC, worker has about 2 screens, so depending on the worker preferences, they may have it on the other screen/session from their production Maxis screen.  If we're to do that, maybe a text message informing them to open it on the same window/session would be good, too?  What you think?

RobertFewins-Kalb:
I think I have found the solution. Methinks the script was trying to navigate back to MAXIS too quickly. I think an EMWaitReady has fixed the problem. Testing the solution.

Y-Lor-Y:
Let me know if you need me to run it after you've made the changes.

RobertFewins-Kalb:
You will need to run this as you normally would, first executing the text in your global variables file, then executing the FuncLib, then executing the script (because of the variable "worker_county_code")...

[...]

PR #979, opened 2015-08-07
Update BULK - FIND MAEPD MEDI CEI.vbs

RobertFewins-Kalb:
Added fix to MMIS navigation logic to always get back to RKEY.


RobertFewins-Kalb:
Closes issue #868

theVKC:
@RobertFewins-Kalb, I'm not sure about this one in general. Mainly, because I've seen lots of folks don't use "C3NN" at all. Me, for example: I get no such selector, because my MMIS access is set to directly go into the recipient file application screen.

That having been said, this may be better than nothing. Let's see how it works, and we'll adjust in beta.

Issue #875, opened 2015-07-20
MnSure retro case note template


IlseFerris:
Mnsure retro processing in MAXIS.  Case note templates are in OneSource.

khiestand:
@IlseFerris if this is just pretty much a simple case noting template script, I think I can probably tackle it!  I was on OneSource looking for the case note templates, could you point me in the right direction?

IlseFerris:
@Y-Lor-Y started working on this, but he had said that anyone can go and enhance it.  I will email you the information that I think is useful.  Thanks @khiestand!

khiestand:
@Y-Lor-Y , do you need any help with this?  I looked at it and I think it looks great!

Y-Lor-Y:
I thought Robert found a functionality or something within the script that is causing this to not work as intended?  I tested it initially and it was able to case note.  Based on the idea from Ilse, it was going to be doing more than just case noting?  At least that was the thought I got when I first read the idea behind this script project, since I ran out of time I thought I put this up for others to enchance it further.

Y-Lor-Y:
Here's what Robert noted:

•  object spacing in the dialog
 •The TIKL check is not connected to logic to TIKL

khiestand:
@Y-Lor-Y , I think I actually know how to fix these things!  I will work on it!

theVKC:
I removed the source file from GitHub so as not to interfere with plans we have to dynamically create dialog files. If either of you require @Y-Lor-Y's file, I can email you a copy. :)

@Y-Lor-Y, because you changed your GitHub account name, you didn't get credit for your original contribution. :/ I may ask you to submit it again, but only after @khiestand is ready to install her fixes. So, we'll do this:

1. Kelly installs her fixes on her copy of the file, tests it, and confirms it's working.
2. Yang navigates to the master repository, and starts a pull request for his incomplete file.
3. I pull that one in, and tag Kelly.
4. Kelly immediately navigates to the master repository, and edits that new file (which initiates a pull request).
5. I approve that file, which will credit both of you.

Does this sound do-able, @khiestand and @Y-Lor-Y?

Y-Lor-Y:
Thank you, All.

khiestand:
@theVKC , I think I do need the file again.  Will you send it to me?

PR #988, opened 2015-08-11
NOTES - MNSURE RETRO HC APPLICATION.vbs

khiestand:
This is a case noting template that details the information about a MNSure retro health care request that a financial worker must process in MAXIS/MMIS. The script will note the following details: curam case number, app rec'd date, gap month, number of retro months requested, if hc closed within 120 days, who coverage is requested for, hc app status, verifs needed, other comments, actions taken, if a tikl was created and if MMIS was updated.  The script then navigates to DAIL/WRIT and puts a date 10 days out for the worker to then type what they requested and any other notes there.

Issue #875

Thanks!


theVKC:
This resolves #875. :)

PR #989, opened 2015-08-11
added Aitkin Co to the party

IlseFerris:
Also updated the FuncLib path (Op Doubtfire)
Upated Check for MAXIS TRUE to FALSE
added Aitkin county to be an 'autofill county'


IlseFerris:
Resolves issue #990

Issue #1004, opened 2015-08-14
BILS updater bug - not updating BILS


courtrightd:
Script appears to simply stop after entering new bills in the dialog and pressing ok.  Pretty sure I found the issue (panel not staying in edit mode), working on fix.

theVKC:
Didn't we fix that? @IlseFerris?

CDPotter:
Kirsten reported that issue this week as well. I hadn't had a chance to dig in quite yet.

theVKC:
@IlseFerris, could this have anything at all to do with that issue you found yesterday with the order of the MAXIS_background_check being incorrect?

IlseFerris:
This very well could be. I pulled in an updated version of the BILS Updater in with Op Doubtfire

theVKC:
Hmmm. This is a critical bug. Can you pull the version out of Doubtfire and pull request separately?

IlseFerris:
I'm on it!

theVKC:
@courtrightd, can you hold off until Ilse pulls her work in?

courtrightd:
Yep – I will wait.  The issue does seem to be from the second CHECK For MAXIS,  I have a temp fix working here by just adding a PF9 before it starts the loops to edit the panel.

IlseFerris:
I pulled it into all 3 branches.  The issue I had yesterday in testing was the MAXIS_background_check prior to STAT/BUDG.  I moved it, and no issues since.

PR #1007, opened 2015-08-14
bug fix issue #1004

IlseFerris:
pulled into release


theVKC:
@IlseFerris, I'm closing this without pulling because this issue was only affecting BETA as far as I can tell. The change for RELEASE affects far more code than the change for BETA, apparently. It looks like you added some extra logic which is currently being tested, and I don't want to move it into production before it's ready. Considering it's 3:50 on Friday and the release is happening Monday, I'm going to let it go for the time being.

IlseFerris:
Sounds like a plan.  You know me, I can't just fix a bug, I've got to mess with other stuff too!

Issue #1008, opened 2015-08-14
LTC MA approval bug


IlseFerris:
if  case is S or B budget it is not case noting the header.  Also not navigating to the correct footer month.

PR #1011, opened 2015-08-14
Bug fix for issue #1008

IlseFerris:
* updated script to navigate to the correct ELIG/HC footer month and
year
* Error handling: if case is not L, S,  or B budget, script will end
with message informing the worker to use the Approved programs script
instead of LTC MA approval script
* Updated approval dialog to update with "$0" if recip amt, income or
deductions do not exist
* updated to have worker_signature to be mandatory field
* updated PAPERLESS IR to Paperless IR (which was causing S and B
budgets to not have a case note header.


theVKC:
Same as the other one. I'm going to defer this one until the beta release goes live Monday morning.

Issue #975, opened 2015-08-07
NEW SCRIPT IDEA >>> BULK script for identifying panel update date on key panels


RobertFewins-Kalb:
New script idea emerged today -- a BULK script that allows the user to select specific caseloads, identify specific panels to check, and a time period criteria to check.

The time period criteria would be something like...
* Identify panels that have been updated within the past 30 days.
* Identify panels that have been updated within the past 6 months (180 days).
* Identify panels that are active but have not been updated in 12 months or more (365 days).
* Identify panels that are active but have not been updated in 24 months or more (730 days).

The available panels would be those focused on income, expenses and deductions.

This script would not be designed to be a tool for financial workers, but instead a supervisory or review tool.

RobertFewins-Kalb:
What luck! @CDPotter  and I have begun work on just such a script!!

IlseFerris:
Your timing is impeccable then :)

Farleight:
Done right, that actually could be a tool for workers to help verify mass changes went through properly (since, ya' know... they usually don't)... just sayin'.

theVKC:
I assigned @CDPotter. If it's a joint effort, please try and cross-pull from a central branch, so as to assign credit to both of you. It may not be important to you, but it is to me and my superiors. :)

PR #1043, opened 2015-08-24
Patch 6

CDPotter:
Script will run through a case load, or multiple case loads and pull the update date off of specific error prone panels selected by reviewer.

#975


theVKC:
Looks good!! Resolves #975

PR #1046, opened 2015-08-26
ACTIONS - TRANSFER CASE

GayS:
added the word "Success!" to the script_end_procedure for statistical tracking purposes.


GayS:
added "Success!" to script end to assist with data tracking report

PR #1061, opened 2015-09-02
critical bug fix in master branch only

IlseFerris:
Duplicate dialogs were in the script (only 1 needed).
Old maxis checking functions removed from DO LOOP


IlseFerris:
For some reason in the OP DOUBTFIRE updates this one got corrupted somehow.

PR #1095, opened 2015-09-14
Update NAV - POLI-TEMP.vbs

Y-Lor-Y:
Alright, sorry for the many requests...this should be last one, based on our conversation.

### DIFF
--- a/Script Files/NAV/NAV - POLI-TEMP.vbs
+++ b/Script Files/NAV/NAV - POLI-TEMP.vbs
@@ -47,14 +47,14 @@ END IF
 'END FUNCTIONS LIBRARY BLOCK================================================================================================
 
 'DIALOGS--------------------------------------------------
-BeginDialog POLI_TEMP_dialog, 0, 0, 256, 60, "POLI/TEMP dialog"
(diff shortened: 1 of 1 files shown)


theVKC:
Sorry @Y-Lor-Y, this doesn't work. You included an extra space after the `Temp_table_index` value on  row 52.

Please test ALL options made when submitting enhancements to scripts, @Y-Lor-Y. Make sure to test them in both dialog editor, as well as Blue Zone Script Host. You should have tested this by:

1. First running it and making sure cancel works (which it does, but ALWAYS test that).
2. Then, running it for the TABLE option (which works)
3. Then, running it for the INDEX option (which doesn't work).

It's very, very important to test all script changes thoroughly prior to submitting them. When testing, it's advisable to:

1. Save a copy of the file somewhere on your flash drive or computer where you'll remember where it is.
2. In BlueZone, navigate to that file using Script>Play.
3. Run the script under every conceivable IF...THEN scenario which is affected by your change.

Please let me know if you have any additional questions. I will be closing your three pull requests, but will be happy to reconsider this script after you've made the above change and tested it.

Issue #1097, opened 2015-09-15
NOTES - HC Renewal & HCAPP footer month


GayS:
Staff are reporting that the case note scripts for HC Renewal & HCAPP  have stopped pulling in the footer month at the start of the script.

CDPotter:
I can confirm I've had this issue here at Anoka as well (both MASTER and BETA branches). Strangely I was looking at the code for the HCAPP and when run through bluezone directly it filled in the month.

CDPotter:
So been doing a little testing this afternoon here are my results.

code in CDP branch is identical to DHS MASTER

Open notes on Power pad->NOTES beta & master-> hcapp = won't fill footer_month
play button-> hcapp(from master) on desktop = fills footer_month
play button -> redirect pointing to hcapp on CDP branch = fills footer_month
play button -> redirect pointing to notes menu on cdp branch -> hcapp (cdp branch) = fills footer_month
play button -> redirect pointing to HCAPP on master branch = won't fill footer_month

I'm at a bit of a loss for what to try...

GayS:
that is really weird - do you know if the footer_month autofill ever worked from the master branch?  Can the file/code be copied/pasted from an earlier version?

CDPotter:
It seemed to be working last month as I recall. I did find this frustrating in testing this morning. When I define footer_month as "10" the line before the do it seems to work. I've asked Robert to take a peak at it since maybe I've been looking at it too long I'm missing something. 

![hcapp](https://cloud.githubusercontent.com/assets/7841989/9934709/9a4b47ee-5d18-11e5-8f4b-a0b7f3ba0ea2.png)

GayS:
Thanks for tackling this one!

RoyStearns:
Since I have no allotted script time I will make this short and sweet- I wonder if this is related to the fix on issue 561?

PR #1102, opened 2015-09-17
Fix for footer month issue

CDPotter:
#1097


CDPotter:
Thanks to @RobertFewins-Kalb for remembering a solution to this that we used when this happened last year. It seems to work on my live branch. I'd like to have it tested for a bit since it isn't critical.

PR #1103, opened 2015-09-17
Updated verbiage on the SNAP E & T letter description

IlseFerris:
from MEMO to SPEC/LETR.  I can't tell you how many people report that the script isn't working because they go to SPEC/MEMO instead of SPEC/WCOM to see the correspondence.


theVKC:
@IlseFerris, can you create an issue for this so that we can close the issue for this?

PR #1104, opened 2015-09-18
SNAP E & T multiple enhancements.

IlseFerris:
*updated DO LOOP with new err_msg standard
*added error handling to prevent users from selecting a date in the past
* made the AM_PM droplist a mandatory field
* changed script to end after saving the SPEC/LETR
* updated the message box  at the end of the script informing the user
that the orientation letter can be viewed in SPEC/WCOM


IlseFerris:
resolves issue #1105

Issue #981, opened 2015-08-10
POLI/TEMP NAV bug


IlseFerris:
If you select "INDEX OF TOPICS", script navigates you to POIL/TEMP - INDEX
If you select "TABLE OF CONTENTS", script navigates you to POLI/TEMP - INDEX

theVKC:
Do you want it?

IlseFerris:
I can take it, unless one of the newer scriptwriters wants to take a crack at it.

Y-Lor-Y:
I've updated the script with drop list rather than radio.  I've tested it and it works to navigate to TEMP/INDEX and TEMP/TABLE.

PR #1112, opened 2015-09-22
Update NAV - POLI-TEMP.vbs

Y-Lor-Y:
I've edited it a little bit, tested, changed the dialog box so that I'm not complicating it as much as I need it to be (missing values & etc).  This should allow the user to navigate to Index and TABLE both ways; before you have to exit out back to SELF manually to use the script again.


theVKC:
Looks great! @Y-Lor-Y, when you submit a pull request, please also tag the issue it resolves. So, this resolves #981, so I will include a number-sign, then 981.

PR #1123, opened 2015-09-24
Patch to PND2 list

RobertFewins-Kalb:
The script is writing the entire 7-digit worker number when running for the entire agency.


theVKC:
@RobertFewins-Kalb, did you test this fix on agency-wide AND on single user AND on multiple user (not agency wide)?

Also: can you manually change your county to another and try it there? Some agencies manage multiple x1## setups, and I want to make sure your fix works with cases where user is one county but global variables is another (as is the case often in multi-county setups)

theVKC:
Also @RobertFewins-Kalb, if this is a fix in master, since we're testing beta fixes too, can you pull it there as well?

RobertFewins-Kalb:
Will do

And it works for both agency-wide and individual cases

Issue #561, opened 2015-04-13
Bug- Case note dialog does not change footer month forward


RoyStearns:
These are the scripts that I know of that do this (there may be others)- CAF, CSR, HC renewal, HCAPP.
If you are trying to use the Case/NOTE scripts above and are in the footer month of 5 and change it to 4 from inside the script
![capture footer](https://cloud.githubusercontent.com/assets/10159949/7123493/05971d72-e1e8-11e4-80c4-396b305bfe99.PNG)
- the script will f3 out and start reading Maxis stats for the case note in the footer month of 4. However if you happen to start the script in the footer month of 4 and want to change it to 5 from inside the script- it will still pull all of the info for the case note from the footer month of 4, not changing to 5.
This has caused confusion as to where the script is getting the info from and also causing incorrect case notes if not caught.

theVKC:
Hmmm. The logic for footer month entry is ancient. As in it was a very early development in my scriptwriter career at Anoka. Do you have ideas for fixing it, @RoyStearns?

theVKC:
I'm going to try and pound this out before vacation this month!

theVKC:
@IlseFerris just built a new function for this stuff. Perhaps after it's been tested we can evaluate again. Maybe in a couple of weeks we can close this issue.

theVKC:
@IlseFerris/@RoyStearns, do either of you know if the new function resolved this?

PR #1134, opened 2015-09-25
Fix for footer month issue

CDPotter:
#561  CAF seems to be pulling footer months, CSR also autofills, as does CAR. HCAPP was fixed in previous version.


theVKC:
Two things, @CDPotter :

1. Do you mean #561?
2. Does this resolve that issue?
3. Can you also pull to BETA if it does?

CDPotter:
I do, yes, and yes.

PR #1136, opened 2015-09-25
Fixes #1131

### LINKED ISSUE #1131: SHELTER FORM RECEIVED - Dialog out of range error.
Happens whenever it is run from a menu (locally from agency custom or when running from github.)  Works when run directly (script->play).  That is as far as I got.

courtrightd:


theVKC:
@courtrightd, can you pull into BETA please?

PR #1152, opened 2015-09-29
Simplifying auto-fill of benefit year

RobertFewins-Kalb:
The script was subtracting 2000 from the benefit year when all it needed to do was generate the right 2 of the benefit year.


theVKC:
duplicate.

PR #1165, opened 2015-09-30

IlseFerris:
variable has wrong name.  Pulling into master.


IlseFerris:
issue #1168 resolves this

PR #1166, opened 2015-09-30

IlseFerris:
variable name incorrect.  Fixed in Release.


IlseFerris:
issue #1168 resolves this

PR #1167, opened 2015-09-30

IlseFerris:
variable had wrong name.  Pulled into BETA.


IlseFerris:
issue #1168 resolves this

PR #1195, opened 2015-10-01
Update NOTES - GOOD CAUSE CLAIMED.vbs

StormageddonDLOA:
Added an additional DHS doc number to the dialog box per request from my supervisor.

This is from issue # 1157.


theVKC:
@StormageddonDLOA, there seems to be a small glitch in your dialog editor. Or it could be a version difference in BlueZone. Your script looks good, but you may notice that your elements are sized to be 16px tall instead of 15px. I worry this could introduce an unstable element from an accessibility standpoint.

Can you check the snap settings in your dialog editor and see that they snap on 5 px instead of another number? If that does not help, check your system settings to make sure you have text set to display at 100% instead of 125% or another number. That has caused goofy issues in the past.

theVKC:
@StormageddonDLOA, I'm closing this one because the dialog size is all wonky. Feel free to re-open it once we get that taken care of. If it isn't today, I'll ask you to wait until hackathon next Monday. :)

PR #1196, opened 2015-10-01
Update NOTES - VERIFICATIONS NEEDED.vbs

StormageddonDLOA:
This update fixes the script not including the case worker name in the case note.

Issue # 1158.

The TIKL part of the script is working as intended. It brings the worker to where they can enter their TIKL. The msgbox tells the worker what they are supposed to do at this point.


theVKC:
This looks good, @StormageddonDLOA! My only request is that you put the `#` symbol next to your number when you write these. That way it creates a dynamic link.

PR #1198, opened 2015-10-01
Create NOTES - IEVS NOTICE RECEIVED.vbs

StormageddonDLOA:
This script creates a case note for IEVS matches. There is a separate dialog box if the IEVS match is resolved, and if the code is non cooperation then there is also a message box that pops up at the end to remind the user to add a DISQ  panel.

This is issue 1197.

This fixes the issues with the original script: capitalizes NOTES & NOTICE, 
    adds a comment in the code to explain why "**DO NOT ENTER ANY OTHER INFO**" is part of the case note, 
    converted part of the case note script to "write_variable_in_CASE_NOTE" from what had been there, 
    cleaned up some "DO loops" and replaced them with err_msg logic, 
    changed the droplist boxes from "Select" to "Select one".


theVKC:
@StormageddonDLOA, the file name is all wonky. It needs a `.vbs` at the end, and needs spaces after the dash. It should read `NOTES - IEVS NOTICE RECEIVED.vbs` for the title. If you are unable to get this done today, that's fine: but it won't be included in tonight's beta release. No big deal since we can do another one after the Stearns hackathon. :)

Issue #1209, opened 2015-10-05
Adding Certain pops app to HC app drop down menu


TMBohm:
CAF script - Could DHS 3876 be added to the list of Forms from the dropdown at :  If HC applied for (or re-certifying) ...?
The title of DHS 3876 could be shortened to DHS3876 Certain Pops.

IlseFerris:
If a client has a recert for HC, then the NOTES - HC renewal should be used.

TMBohm:
Thanks, Ilse.  Do you know why that option is on the CAF script?

Treva Bohm
Eligibility Worker
218.470.8452

IlseFerris:
Good question.  I think that is for cases that have both HC, and other programs.  Best practices question for the group.  Should users run both the CAF and HC renewal if they are open on HC and other programs? @CDPotter  @RobertFewins-Kalb @courtrightd

CDPotter:
Personally if I have a CAF renewal and a HC renewal I just run the CAF and use that drop down. I've never had a situation where someone applied with certain pops and a CAF though but I would think you can use it the same way. It's an editable box so you can enter any document you receive. 

I would think we could add certian pops to it.

TMBohm:
We see quite a few Certain Pops Applications, so it would be nice to have Certain Pops added.

Thanks, Everyone!

Treva Bohm
Eligibility Worker
218.470.8452

IlseFerris:
Ok, who wants the issue then?

CDPotter:
I can knock it out easily enough.

TMBohm:
Thanks, Charles.  I appreciate that.

Treva Bohm
Eligibility Worker
218.470.8452

theVKC:
Resolved by #1212. @TMBohm, it'll be included in the next release. Thanks @CDPotter!

PR #1212, opened 2015-10-05
Adding Certain pops app to HC app drop down menu

CDPotter:
#1209


CDPotter:
BLIP: Adding Certain populations application as option on HC Application drop down menu.

Issue #1213, opened 2015-10-05
Bulk Lists may not working with some multicounty agencies. 


RobertFewins-Kalb:
Worry not. I am on it.

RobertFewins-Kalb:
#IGTS

theVKC:
I added these three, @RobertFewins-Kalb. A couple of things:

1. Not a critical bug. Only affects a few people statewide.
2. Any more scripts affected? If so, please pull them to master and beta this week, as this is a bug fix.

theVKC:
@RobertFewins-Kalb, this is done, right?

RobertFewins-Kalb:
I believe so. Looks like I updated BULK - INAC LIST as well, and I'm pretty sure the other BULK list scripts that needed multi-county handling already had it.

theVKC:
Thanks @RobertFewins-Kalb!

PR #1215, opened 2015-10-05
reinstating multicounty functionality

RobertFewins-Kalb:
this script would not work for multicounty agencies. now it should. I inserted the function that does stuff to reduce multicounty to a single 2-digimon county code


RobertFewins-Kalb:
addresses issue #1213

PR #1216, opened 2015-10-05
reinstating multicounty functionality

RobertFewins-Kalb:
multicounty handling was not enabled. the function has been reinserted so multicounty agencies can has script


RobertFewins-Kalb:
addresses issue #1213

Issue #1220, opened 2015-10-07
BUG -- NOTES - APPROVED PROGS -- the script is not set-up to select the appropriate notice to write to


RobertFewins-Kalb:
I have referenced this bug in #1217

theVKC:
Resolved by #1227.

Issue #1217, opened 2015-10-05
Notes - Approved Programs - CS disregard wcom may have an old date


Y-Lor-Y:
The CS disregard fiated wcom note to client has the earlier release date of 07/01/2015.  So that we're align to when to start this disregard, DHS puts this back on the table effective 10/01/2015.  Should this date be changed to 10/01/2015 for the WCOM to not confuse the client?

I can take this one on.

theVKC:
Anyone have any thoughts on this?

courtrightd:
Yeah – it definitely has an old date, and I’m worried we may have clients rather unhappy if they start receiving disregard in Oct.  but the notice says it was effective 7/1/15.

theVKC:
@Y-Lor-Y, if you want to take this one on, be my guest. If you could pull it into both master and BETA, that'd be swell. I'm going to tag this as a bug, because the date is merky and probably considered incorrect.

RobertFewins-Kalb:
Additional problem!!!

The script is set up to write to a notice on WCOM row 7. But we originally built the WCOM to include a DO/LOOP to find the appropriate notice. 

Whoever does this needs to make sure they are building a do...loop around this...
![image](https://cloud.githubusercontent.com/assets/7842016/10342398/823ee162-6cde-11e5-9c8e-02a067a7bf00.png)


...to make sure the script is finding the appropriate notice.

RobertFewins-Kalb:
@Y-Lor-Y , can I tackle both issues? I know exactly what needs to get done.

Y-Lor-Y:
@RobertFewins-Kalb @theVKC  I'm okay with that.  Take it on!  Also, I've just noticed, ontop of the date change:  row 470, that's where the case note to the spec/wcom - the date change from July to October.
Not that this is an error, but for the display of the case note, row 548, col 72, has an extra space.
There is also another minor typo:  Row 548, Col: 108,  "as" should be "has" right?

RobertFewins-Kalb:
Also, APPROVED PROGRAMS is getting the Operation Wind Tunnel treatment for err_msg handling.

theVKC:
Resolved by #1227.

PR #1228, opened 2015-10-08
several fixes to NOTES-APPROVED PROGRAMS (see below)

RobertFewins-Kalb:
This update includes the following fixes, addressing issues #1217 and #1220.
1. err_msg handling around dialog (Operation Wind Tunnel)
2. Updated verbiage in notice to reflect October start date for CS Disregard (rather than July) and also removing typo.
3. Added do...loop around search for appropriate WCOM when user selects to case note that the CS disregard verbiage should be added to notice.
4. Updated case note verbiage to case note only when script actually updates a notice (currently, it is case noting that the cs disregard verbiage was added to the notice even if the script did not add one...there are instances when a new notice is not generated).


RobertFewins-Kalb:
needs to go into BETA. sorry

PR #1240, opened 2015-10-12
Fixed a few issues that were holding some scenarios up.

CDPotter:


theVKC:
@CDPotter, this isn't going to hold up my acceptance of this, but can you tell me why there's a FuncLib call commented out at the top now?

CDPotter:
I think we were testing on a local version of func lib. I don't think we were changing any of the functions though so I can't think why. I'd wager it was left over from previous testing since we hadn't uploaded our newer version since a few updates. @RobertFewins-Kalb and I really need to clean up our files.

RobertFewins-Kalb:
FuncLib needs a few updates for Krabappel to work. Rather than hitting @theVKC with a couple more pull requests, we just made the updates locally to get Krabappel through.

We can submit pull requests to get the functions updated. They are exclusively Krabappel functions, so it's not like we've been holding out on releasing fixes that would improve the rest of the scripts.

Issue #1149, opened 2015-09-28
Notes - MFIP Sanction 


Y-Lor-Y:
The dialog box verbiage indicate that the sanction needs to be cured by 10 days cut-off date, however, per policy it's actually 10 days from the effective month.  So, end of month minus 10.  For more detail, refer to PQ # 9892 (1/21/11) for further references.

theVKC:
Are you volunteering to take this, @Y-Lor-Y?

Y-Lor-Y:
Sure, I'll take a stab at it.  Note my time are limited, but if this is not a rush I would like to.

PR #1245, opened 2015-10-12
Update NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs

Y-Lor-Y:
Resolving issue #1149  The verbiage and date of compliance needed to be update.  I've added for it to compute the resolution/compliance date to be display within the case note.

### DIFF
--- a/Script Files/NOTES/NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs
+++ b/Script Files/NOTES/NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs
@@ -83,9 +83,7 @@ END IF
 'DIM worker_signature
 'DIM ButtonPressed
 'DIM TIKL_date
-
-
-
+DIM Resolution_date
 
 'DIALOGS----------------------------------------------------------------------------------------------------
 'MFIP Sanction/DWP Disqualification Dialog Box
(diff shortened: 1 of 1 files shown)


theVKC:
@Y-Lor-Y, I'm going to close this without accepting it. I see a few glitchy things on it:
1. You added a space between "7" and "+" on `number_occurances_droplist` in the dialog, without adjusting the results of this elsewhere, or explaining why it was done in your update.
2. It looks like you removed the `last_day_cure` editbox, but I don't see a corresponding `resolution_date` editbox. I'm not sure from your notes here why this was removed, or why `resolution_date` isn't an editbox itself.
3. You appear to have DIM'd `resolution_date` without commenting it out. DIMing is good, but we comment those out when we commit (just to avoid any possibility of another DIM with the same name taking place somewhere in the chain of scripts leading to this one, you never know). No big deal (just comment it out).

Feel free to correct the above and resubmit (although if it's after today, let's wait until next week's release, as I'll be out of the office for the rest of the week).

Y-Lor-Y:
#1)  That was my fault.  I noticed there were two 7 7 values, and from my understanding the 2nd 7 would’ve been sanctions beyond the 1st 7th time.  I thought I added it correctly in the dialog boxes update, but guess it didn’t as I thought.  I will change that.

#2) I no longer need the last_cure_date editbox.  Because it did not make sense to have it there for worker put the effective date.  IE:  Sanctions dates are always 10 days or 1 days from the effective date.  So you’d take the effective month minus 10 or 1, that is the resolution date.  If I could have where, that math can be populated into the dialog box, I could set up the dialog to do that filler in it.  Is there an ‘autofill’ option?  So that once resolution date is known, it’d show that on the dialogbox.  (I asked you this when we were in St. Cloud).  From that, seems that’d be a no.

I removed the editbox and instead put in a text box to further explained to workers that the date is in the case note.

#3) The reason why I DIM’ed Resolution_date, when my:

IF (sanction_type_droplist = "ES") or (sanction_type_droplist = "No show to orientation") or (sanction_type_droplist = "Minor mom truancy") then
                Resolution_date = DateAdd("d", -10, Date_sanction)
ELSEIf (sanction_type_droplist = "CS") then
                Resolution_date = DateAdd("d", -1, Date_sanction)
End If

It has a placeholder, so that when it do the case notes:
CALL write_bullet_and_variable_in_case_note("Last day to cure", Resolution_date)

It has a placeholder to refer to the variable.

In the above situation was a DIM not necessary and I’m creating more issue?  ☺
(in my edit files, I now just updated the explanations)


NEW:

#4)  I’m thinking more, and that I just thought of another issue that needs to be add while we’re at this.
A check box that if it’s check for consecutive sanctions month, that the resolution date = effective month minus 1 (quick add).

#5) I want to add to it another check box that if it’s check if the Sanctioned approval has been approved  Or Spec/LETR has been sent
       For it to read the info and do a spec/wcom notes with ES info, effective sanctions by.  **caution** this gets tricky once beyond the 7th sanctions.  How would I have a dialog box for workers to put in the sanction amount of time?  Should I changed the dropdown list and change that to be an edit box so that the worker can just put in a # in there.  It’s more confusing to have it as 7+  vs. 8 or 9 or 10 or 15.


Your thoughts?

Yang T. Lor l FCM Resource Trainer
Financial Assistance Services l  Ramsey County CHS
160 Kellogg Boulevard East, Saint Paul, MN  55101
P: 651-266-3621  l  Fax: 651-266-3705
E: yang.tou.lor@co.ramsey.mn.us
Schedule: M to Th, 7:45 AM to 6:00 PM

The information included in this email message, including any attachments, is intended only for the person or organization to which it is addressed.  This email message may contain information that is privileged or confidential.  If you receive this email message and are not the intended recipient or responsible for delivering the message to the intended recipient, you may not use, disseminate, distribute, or copy the information included in this email and any attachments.  If you received this email by mistake, please reply by email and destroy all copies of this message and any attachments.  Thank you

PR #1269, opened 2015-10-15
Fixed bug when canceling script & footer month

IlseFerris:
Bugs fixed include footer month not filling itself in, and when user
would push the cancel button  the case was going through MAXIS
navigation.  Also cleaned up a couple functions.


IlseFerris:
Resolves issue #1270 
Blip for release notes: Bugs fixed include footer month not filling itself in, and allowing the user to cancel without navigation through MAXIS panels before hand.

Issue #1273, opened 2015-10-15
NOTES - Expedited Screening is no longer providing a msgbox of results when it can't case note


RobertFewins-Kalb:
@CDPotter had fixed this, but I believe doubtfire reduced the script's functionality.

CDPotter:
I'm looking into a fix.

theVKC:
"Fixed" = enhancement. #ImNotStillBitterAtAll

Also, I think a better, more permanent solution to our woes would be to modify the function itself to provide some sort of doublecheck like this. That way, this sort of thing would be detected.

I think it also speaks to a need for scriptwriters to consider testing multiple scenarios when working on enhancements. Had we tested an out-of-county case, we would have discovered this.

PR #1276, opened 2015-10-15
Bug fix/reverting code

CDPotter:
Op Doubtfire replaced some code that was in place to prevent an infinite loop that start_a_blank_case_note causes when someone uses an incorrect case number or out of county case number. Took code from #967 and reincorporated it with recent enhancements.

BLIP: Expedited screening tool can once again be used on out of county cases or cases without a case number (enter 1 as case number if client doesn't have case.)

fixes issue #1273


theVKC:
Closing this one, as there'll be a beta release today and I want to avoid merge conflicts.

PR #1293, opened 2015-10-21
Updating WF1 Referral in Krabappel

RobertFewins-Kalb:
Now with extra strength WF1 referral handling.


theVKC:
@RobertFewins-Kalb, was there an issue that this resolves? If not, can you quick make one and tag this pull request in it? #Tracking

Issue #1300, opened 2015-10-22
Enhancement CSR Next button


CDPotter:
CSR currently has an OK button on first panel. I'm thinking it might be a bit more clear if it was a next instead of an OK similar to the CAF. Submitting fix.

PR #1301, opened 2015-10-22
Changing OK button to NEXT

CDPotter:
BLIP: Changing OK button to a next on the first dialog to help the user see there is more dialogs to come. 

#1300


theVKC:
@CDPotter, friendly reminder: the week after a beta release is a script freeze. It's not so big of a deal because I wasn't working, but just know that the week after the beta release should only see pull requests for bug fixes.

Issue #1302, opened 2015-10-23
BUG Verifs needed checkboxes don't do anything


CDPotter:
Reported on SIR:

When box is checked for "DHS2919A -Verification A is sent" it is not carrying over into the case note.

Another issue with this script is it's not carrying over the Sent to AREP option to the case/note. Thought I'd mention that on this thread rather than creating a new thread since it's for the same script.

Submitting fix

theVKC:
Resolved by #1303. Not sure why it didn't close.

PR #1304, opened 2015-10-23
Adding checkbox functionality

CDPotter:
BLIP: some checkboxes were not working, readded functionality that was lost at some point. 

#1302


theVKC:
@CDPotter, I'm closing this, but only because I'm about to issue the release, and it's already included in the master branch. All branches will be in sync after this happens, so there's no need for separate pull requests (this is only because I was out and the release is due today).

Issue #1263, opened 2015-10-13
Enhancement for Duplicate Assistance WCOM - available for MFIP


C-Love:
Can we add functionality for this WCOM to work on an MFIP case when there is a duplicate assistance situation?

theVKC:
Can we? Sure!
Should we? Do you feel like doing it, @C-Love?

C-Love:
Yes, I can will take it.

PR #1309, opened 2015-10-26
Update MEMOS - DUPLICATE ASSISTANCE WCOM.vbs

C-Love:
Script will now add a WCOM to an MFIP notice as well - when client on case has received SNAP on another case.

### DIFF
--- a/Script Files/MEMOS/MEMOS - DUPLICATE ASSISTANCE WCOM.vbs
+++ b/Script Files/MEMOS/MEMOS - DUPLICATE ASSISTANCE WCOM.vbs
@@ -63,7 +63,7 @@ EndDialog
 EMConnect ""
 
 'warning box
-Msgbox "Warning: If you have multiple waiting SNAP results this script may be unable to find the most recent one. Please process manually in those instances." & vbNewLine & vbNewLine &_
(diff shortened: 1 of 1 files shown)


C-Love:
Issue #1263

theVKC:
Hey @C-Love, just noticed this: in the WCOM when it adds the notice, you're using `else if` instead of `elseif`. I have a couple of concerns about that:

1. `Else if` is somewhat poor syntax in VBscript, it behaves different than `elseif` from my experience (https://msdn.microsoft.com/en-us/library/5h27x7e9(v=vs.84).aspx)
2.  Is it either an MFIP notice or a SNAP notice? i.e., can't be both at once? If it can be both simultaneously, it should be separate IF...THEN statements, not IF...ELSEIF statements. Make sense?

theVKC:
@C-Love, I'm closing this pull request just to keep things clean. Please repull when you get the above resolved (if you did that already, thank you). :)

C-Love:
Thank you. Sounds good - I think I have it done - just needs a bit more testing.

theVKC:
Good deal! Enhancement freeze ends Monday, at which point feel free to resubmit. :)

PR #1317, opened 2015-10-28
Update ACTIONS - CHECK EDRS.vbs

MelissaFox-Stearns:
The variable name SSN_Number was incorrectly spelled SS_Number.  (Whoops!) 
I made the fix and it appears to be working a-ok now.


theVKC:
@MelissaFox-Stearns, can you make an issue for this? Also, if this is live for everyone, can you please also pull it into the BETA and RELEASE branches? (Email if you want help with that).

PR #1326, opened 2015-10-29
Update ACTIONS - CS DISREGARD FIAT.vbs

### LINKED ISSUE #1321: Action - CS Disregard Fiater
I was running this script today and it did not seem to be working or something is off.
What happen is that the script would run just fine until it get to the parts where it checks for the UNEA panels - if found, it gives the message box that it found one of the three incomes type, to press ok to continue.  After it went through the UNEA panels checking and pressed OK to continue.  It then skips all the way down to the script end procedure message - showing the message that it is success and to run the approved program.  In actuality, no FIAT'ing was done.

So, the whole part of telling it to go and process the fiat'ing was skipped.

I'm looking at the coding...I noticed two area, not sure if they may be out of place or not.  Or, it could totally be something else.

Row 197, The Loop is not define, I thought Loops has to be define or conditioned.  Maybe not in this case?

Row 242's There is a transmit put on this row, just seems out of place to me.

Thanks,
Yang

courtrightd:
Resolves #1321  Removed <> "APP CLO" from the program checks on CASE CURR, changed logic to look at DWP, and if no version to look at MFIP.  (For when transitioning from DWP to MFIP. This is the only "APP CLOSE" situation I could think of that would cause an error.)  @RobertFewins-Kalb can enlighten us on Monday if there is still an issue.

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - CS DISREGARD FIAT.vbs
+++ b/Script Files/ACTIONS/ACTIONS - CS DISREGARD FIAT.vbs
@@ -250,81 +250,86 @@ IF CS_found <> True THEN script_end_procedure("A child support UNEA panel was no
 back_to_self
 
 'Checking out the sweet, sweet eligibility results, begining with D to the Dubs P
-IF DWP_cash_status <> "" AND DWP_cash_status <> "APP CLO"  Then
+IF DWP_cash_status <> "" THEN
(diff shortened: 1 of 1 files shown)


CDPotter:
That sounds about right, I think there was issues with transitioning between programs.

Issue #1328, opened 2015-10-29
DAIL - NEW HIRE: error out if there's five panels there


theVKC:
It should notify the worker to delete one of those. Right now, it might be causing issues with breaking the functions.

Any takers?

CDPotter:
I can do this.

theVKC:
Thanks @CDPotter!

PR #1331, opened 2015-10-29
Adding check to for 5 jobs panel

CDPotter:
BLIP: Script will now end and prompt user to clean up jobs panels if there are already 5. (MAXIS has a limit of 5 jobs panels per person).

#1328


theVKC:
Not a bug, so I'm closing. It'll go through the regular enhancement process.

Issue #1335, opened 2015-10-30
BULK - DAIL REPORT needs comments


IlseFerris:
Comments, where are you? Trying to learn more about BULK scripts, and could use some comments on this script. @RobertFewins-Kalb can the gauntlet be passed to you?

RobertFewins-Kalb:
I can and shall. Sorry aboot that.

IlseFerris:
Thank you! I'm just trying to reverse engineer and could use the comments.

PR #1336, opened 2015-10-30
Adding comments to BULK-DAIL REPORT.vbs

RobertFewins-Kalb:
I guess I just busted this out without considering that other people would want to know how it works. Sorry, gang.

### DIFF
--- a/Script Files/BULK/BULK - DAIL REPORT.vbs
+++ b/Script Files/BULK/BULK - DAIL REPORT.vbs
@@ -62,6 +62,7 @@ objExcel.Visible = True
 Set objWorkbook = objExcel.Workbooks.Add() 
 objExcel.DisplayAlerts = True
 
+'Excel headers and formatting the columns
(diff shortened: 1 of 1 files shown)


RobertFewins-Kalb:
In response to issue #1335

IlseFerris:
Awesome, thank you!  All is forgiven :)

RobertFewins-Kalb:
Let me know if it needs additional comments

theVKC:
@RobertFewins-Kalb, why'd you set the obj.excel to be not visible?

RobertFewins-Kalb:
I am not sure. I think I was going for effect when the report becomes visible at the end. I'll go ahead and change that...

RobertFewins-Kalb:
It is set to True. I think originally I was playing with the idea of setting it to False so that it would run invisibly and then pop up with the results. I deleted the extra objExcel.Visible = True because it was not needed.

Issue #1268, opened 2015-10-15
HRF- Add case note grant info from Elig/FS for UHFS cases


MelissaFox-Stearns:
Coworker brought up today that the HRF has the option to check the box to case note grant info from Elig/Mfip, however if the HRF is for UHFS, it has no option to case note the elig results from SNAP. 

I think it would be cool if we also made a check box for the SNAP (UHFS) to case note the amounts.

theVKC:
Do you feel like taking this on @MelissaFox-Stearns?

MelissaFox-Stearns:
Yes! Assign away :)

PR #1342, opened 2015-11-02
Update NOTES - HRF.vbs

MelissaFox-Stearns:
Update adds a checkbox to the dialog that gives the option of case noting the elig/fs results for UHFS cases. Issue# 1268


theVKC:
#1268

Issue #1121, opened 2015-09-24
Burial assets enhancement: list in header which program assets were evaluated for


IlseFerris:
Received request for this as the current burial assets scripts gives a breakdown of what is not counted/ counted, but since there are different rules for CASH programs vs. HC, a clear distinction should be made.

IlseFerris:
Also could maybe use a message at the end with the HCPM and CM references, and a reminder that assets excluded for HC may not be excluded for CASH programs.

theVKC:
Do you want this one?

IlseFerris:
Of course!

PR #1343, opened 2015-11-02
Burial assets program header

IlseFerris:
Resolves issue #1121 
 
BLIP: Enhancement to the script that provides a field to enter the program that the burial asset is being evaluated, and fix to the dialog navigation to allow users to navigate back and forth from the first assets dialog though the fourth asset dialog.


IlseFerris:
resolves issue #1121

Issue #1349, opened 2015-11-03
Critial bug SNAP E & T --Hennepin County addr change effective immediately


IlseFerris:
Not a lot of notice on this one, but 2 of the regional offices are moving E & T orientations effective immediately.

PR #1353, opened 2015-11-04
Critial enhancement SNAP E & T HC addr change effective immediately (Release)

IlseFerris:
Century plaza orientations have moved to HSB effective immediately
Updated script to default to MEMB 01.  Added the following Hennepin
County specific features:
* dialog box to reduce errors for scheduling
* Added msgbox with Do Loop that will keep seeking the next available
weekday an orientation is scheduled on until it reaches the desired date
* script end procedure ending message that informs user to send a
specific diamond form

Resolves issue #1349


theVKC:
This was resolved with a pull request from master>beta.

Issue #1354, opened 2015-11-04
Bug when hitting cancel on a DAIL Scrubber Hire Match


StormageddonDLOA:
When ever I hit cancel on a Hire Match when running the DAIL Scrubber the script does some interesting things, that I'm not sure even how to explain, and I have to stop script to get it to stop bugging out.

CDPotter:
I can confirm I ran into this today too when I hit the x button on the new hire message.

theVKC:
Any chance MAXIS is to blame, @CDPotter or @StormageddonDLOA?

CDPotter:
Hard to say, just ran it on another person's dail (cleared mine already) and was able to reproduce it first try. I'd wager it's something with cancel confirmation?

StormageddonDLOA:
Don’t know.

CDPotter:
Most likely doubtfire related. Those new functions require case_number and a lot of the dail scrubbers just hit S on the fail.

theVKC:
But that was over two months ago. We would have seen it last month.

StormageddonDLOA:
Only if someone hit cancel.

RobertFewins-Kalb:
The problem is MAXIS_navigation function. There's no do/loop around the dialog. It looks like the script is trying to navigate_to_MAXIS_screen but it is not working with a case number...

![image](https://cloud.githubusercontent.com/assets/7842016/10943659/e57bb8a6-82da-11e5-883f-f7b1eae45150.png)

theVKC:
But we should have seen this last month. Can someone create a local version and test? I'm still running the QC sample so I'm out of commission.

CDPotter:
Only if someone hit cancel and thought to report it. I added a do loop and reordered cancel_confirmation and maxis_nav and it works perfectly locally. Submitting fix.

RobertFewins-Kalb:
My guess is that user so infrequently cancel the new hire DAIL scrubber script that we haven't had a chance to hit that error.

PR #1355, opened 2015-11-04
Fixing cancel_confirmation bug

CDPotter:
BLIP: Cancelling the script would lead to an infinite loop because of MAXIS_dialog_navigation function. Added a do/loop and reordered functions so they work in tandem. 

#1354

### DIFF
--- a/Script Files/DAIL/DAIL - NEW HIRE.vbs
+++ b/Script Files/DAIL/DAIL - NEW HIRE.vbs
@@ -170,9 +170,11 @@ TIKL_checkbox = checked
 HH_memb_row = 5 
 
 'Show dialog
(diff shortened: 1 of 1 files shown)


theVKC:
@CDPotter, was this tested? I have concerns about changes without testing (your comment doesn't say "tested and it's working" or anything like that, so I feel I should ask...)

CDPotter:
I tested it on at 3 cases so far. Before the cancel confirmation dialog wasn't even popping up. Now it is and it actually cancels the script rather than trying to work through the MAXIS_nav function without a case number.

RobertFewins-Kalb:
...or a valid screen to navigate to.

PR #1356, opened 2015-11-04
Fixing cancel_confirmation bug

CDPotter:
BLIP: Cancelling the script would lead to an infinite loop because of MAXIS_dialog_navigation function. Added a do/loop and reordered functions so they work in tandem.

#1354


theVKC:
This was resolved with a pull request from master>beta.

PR #1362, opened 2015-11-05
Update SETTINGS - GLOBAL VARIABLES.vbs

### LINKED ISSUE #671: GlobVar: add "updated date" variable
That way, FuncLib or other scripts could include a check of that variable to determine if the scripts were updated. This could be helpful for alerting staff to new changes which require re-installation. Not that that happens a lot, but this would be good to incorporate.

theVKC:
This resolves #671. I also created an issue on the installer repo to update this. I set the updated date as the future, because scriptwriters are our future (and it bypasses any planned breakpoints like "you haven't updated your scripts recently enough").


theVKC:
I'll leave this out until after the release. :)

Issue #1366, opened 2015-11-09
NOTES - HRF -- cancel does not cancel the script


RobertFewins-Kalb:
The maxis_navigation and cancel_confirmation function calls are out of order.

IlseFerris:
@RobertFewins-Kalb I can take that if you haven't started already

RobertFewins-Kalb:
I'm already working on it. I'm making a few other updates including err_msg handling.

I would like to make sure we correct this for all the scripts where the problem exists, however. Here are the lines of code in question...
![image](https://cloud.githubusercontent.com/assets/7842016/11041357/f45789e0-86d6-11e5-90d1-ffd645535068.png)

theVKC:
@RobertFewins-Kalb, a release is about to go out. Can I get an update on this?

PR #1368, opened 2015-11-09
Bug fix to maxis_navigation and cancel_confirmation order

RobertFewins-Kalb:
changed the order of the function calls -- it has to be...
1. cancel_confirmation
2. call check_for_maxis(false)
3. maxis_dialog_navigation

Added err_msg handling

Removed superfluous dialog for case note confirmation


theVKC:
Why is this in BETA, @RobertFewins-Kalb? Should go master and BETA.

theVKC:
Also, you didn't tag the issue from the list.

theVKC:
#1366.

PR #1374, opened 2015-11-10
Update NOTES - HRF.vbs

RobertFewins-Kalb:
changed the order of the function calls -- it has to be...
1. cancel_confirmation
2. call check_for_maxis(false)
3. maxis_dialog_navigation

Added err_msg handling

Removed superfluous dialog for case note confirmation


RobertFewins-Kalb:
#1366

PR #1396, opened 2015-11-13
Resolves #1338

### LINKED ISSUE #1338: Locked out of MAXIS.
Bluezone version issues causing workers to get locked out of MAXIS.

MelissaFox-Stearns:
Added new function call check_for_password which resolves Stearns county's problem with Bluezone version 6+ passwording out.


IlseFerris:
@MelissaFox-Stearns for the win!

PR #1398, opened 2015-11-13
data validation enhancement for IEVS match received

### LINKED ISSUE #1397: Enhancement -- IEVS Received needs validation around dialog
A worker reported today an error where the header was IEVS Notice received: Select

It looks like it just needs validation for the value of "OPTIONS"

RobertFewins-Kalb:
variable "OPTIONS" was checking for "Select one" but the DLB option was "Select"

Resolves #1397


RobertFewins-Kalb:
Blip: This adds correction to logic that is checking for mandatory fields. A typo was causing the logic to fail.

Issue #1367, opened 2015-11-09
Enhancement for the Transfer Action script.


StormageddonDLOA:
Add Health Care as an option to the drop down selection box for "Unit to transfer to:" so the case note is correct when transferring a case to a Health Care only unit.

StormageddonDLOA:
I can take this one.

PR #1410, opened 2015-11-18
Update ACTIONS - TRANSFER CASE.vbs

StormageddonDLOA:
HC added as an option to the drop down selection box for unit to transfer to correcting a case noting issue in counties with a HC unit. Issue #1367


theVKC:
Because this is an enhancement and NOT a bug, I can't accept it right now: I will wait until Monday and then pull it in. :)

Issue #1318, opened 2015-10-28
MFIP housing grant work around


IlseFerris:
ACTION script to FIAT the $50 of counted HUD funds out of the MFIP budget (in FMBI?) to ensure that the housing grant is being added to the cases that are exempt from counting the subsidy income.  This would be similar to the Child support FIATer in nature, and maybe we should consider merging these since many MFIP cases need both FIATs.
I have a PDF from mentor meeting today that details out the steps. MAXIS Mentors were informed at this meeting that this will be fixed in MAXIS at some point in 2016, but no date has been given at this point.  I can send this document to anyone who wants it, but could not load it to the issue.

Thoughts on this?

theVKC:
Do you want it?

IlseFerris:
If it's still help wanted on Monday, then yes. I'm going to try to knock out some of my other issues prior to committing.

theVKC:
@StormageddonDLOA has volunteered to take this on. Thank you!

IlseFerris:
New POLI/TEMP reference for this: TE02.10.100 HOUSING GRANT WITH SUBSIDIZED HOUSING

theVKC:
[WORKAROUND FOR SPECIFIC MFIP UNITS RECEIVING SUBSIDIZED HOUSING.pdf](https://github.com/MN-Script-Team/DHS-MAXIS-Scripts/files/25430/WORKAROUND.FOR.SPECIFIC.MFIP.UNITS.RECEIVING.SUBSIDIZED.HOUSING.pdf)

IlseFerris:
@StormageddonDLOA any progress on this one?

StormageddonDLOA:
Yeah I’ve got it done and a couple of my co-workers are testing it. Checked with Veronica on a couple changes I made to bring it more in line with the CS exclusion and got her approval.

IlseFerris:
Awesome!

PR #1414, opened 2015-11-19
Create ACTIONS - HOUSING GRANT FIATER

StormageddonDLOA:
This new script will do the FIATing for the worker when a MFIP case is receiving a subsidy but is still eligible for the Housing Grant. Finishes issue #1318.

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - HOUSING GRANT FIATER
+++ b/Script Files/ACTIONS/ACTIONS - HOUSING GRANT FIATER
@@ -1,0 +1,183 @@
+'Created by Tim DeLong from Stearns County.
+
+'STATS GATHERING----------------------------------------------------------------------------------------------------
+name_of_script = "ACTIONS - Housing Grant FIATer.vbs"
+start_time = timer
+
+'LOADING FUNCTIONS LIBRARY FROM GITHUB REPOSITORY===========================================================================
+IF IsEmpty(FuncLib_URL) = TRUE THEN	'Shouldn't load FuncLib if it already loaded once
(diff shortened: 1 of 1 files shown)


theVKC:
@StormageddonDLOA: closing this issue, not because I don’t love the script, but due to a few nitpicky issues with it. :)
1. The .vbs is missing from the file itself (rename the file).
2. `name_of_script` needs all characters with caps, except for the file extension
3. We have a function for case number finder, and I think I'd like to use it here. In fact, I think you helped with that one. :)
4. Could you add another loop around the main dialog containing our new check_for_password function? See CAF for an example (or ask @MelissaFox-Stearns)
5. When you go to FIAT, you need to go to `("FIAT", "____")`. This clears out any residual data in the command line which may be left over due to MAXIS being weird (it's good error proofing).
6. Needs a `script_end_procedure("")` at the end, for statistics (@GayS would be pleased to include this)
7. I'll need you to add a button to the ACTIONS - MAIN MENU script for this. :)

Otherwise, I love the content. It looks solid to me. The only other issue is we're in an enhancement freeze until Monday so we can test the new enhancements elsewhere (see SIR scripts calendar). So, please pull this again Monday with the changes if you can. :)

StormageddonDLOA:
I’m out next week so I’ll do the pull request the week after.

theVKC:
That should be fine, @StormageddonDLOA. If you can get it in early Monday, I can easily incorporate it into the next beta.

StormageddonDLOA:
I am having an issue where my dialog editor won’t let me make any dimension an odd number.

i.e. if I try to make a text box 60, 55, 215, 10 the editor will change it to 60, 54, 216, 10

theVKC:
This could be that version difference. You could always just find/replace that.

Issue #1429, opened 2015-11-20
Main Menu MEMOS MAEPD no initial premium script not accessable


IlseFerris:
file extension not present so script cannot be accessed. Submiting a fix to all branches.

theVKC:
@MN-Script-Team/county-agency-scriptwriters

A reminder to all folks building and testing scripts: it's vital that when we release new scripts into the master/beta branch for testing (and you-all get a copy of my release notes), that several of us try the new scripts to make sure that at the very least, the buttons work and no fatal flaws are detected.

Lately, I haven't had as much time to test details of functionality. I review the code, and have been relying on beta users to notify us of bugs. This is a pretty big bug that was missed, and it exposes the need for dedicated testing of each script.

I'll start testing the buttons as much as I can, but it will mean I have less time to devote to other areas of scripts, including building newer/better/stronger functions. This critical bug, which has gone undiscovered for weeks, should give us pause as a group. We all need to collectively do a better job of testing new scripts.

PR #1430, opened 2015-11-20
 file extension not included, bug fix in BETA

IlseFerris:
No file extension for MA EPD no initial premium made. Also removed old dates from the dialog boxes.

BLIP: bug fix to enable the script file to be accessed by users deployed, and updated "NEW" labels for new scripts from 10/15 - 11/15.


IlseFerris:
resolves #1429

Issue #1463, opened 2015-11-25
COLA 2016 needs button to be updated on NOTES menu


theVKC:
Can you take this one @IlseFerris?

IlseFerris:
On it!

PR #1466, opened 2015-11-25
Updated COLA 2016 button/critical bug fix issue #1463 RELEASE branch

IlseFerris:
resolves issue #1463


theVKC:
So we have a problem: this contains the new FuncLib call, but RELEASE uses the old one... can you resubmit with the old FuncLib call?

PR #1498, opened 2015-12-03
Hennepin co nomi enhancements

IlseFerris:
Enhancements to the script include: a droplist of times for the missed appointment time, specific office information for Hennepin County users, and specific dialogs for Hennepin County with office information.


theVKC:
@IlseFerris, does this have an issue on GitHub? It should, even if it only affects one county, if it's going into the state repo.

PR #1535, opened 2015-12-10
Fix to footer month & year

### LINKED ISSUE #1534: Bugs >> Footer Month & Year are still wonky in CAF
Footer month is still not showing up (I think it's because of string vs. numeric).

I will submit a fix.

RobertFewins-Kalb:
making both values strings to make sure they are being pulled into the dlg

Resolves #1534 

BLIP: Converting footer_month and footer_year to strings to make sure they are pre-populated in case_number_dialog when it loads.


theVKC:
FYI, that blip, while helpful for me, doesn't really describe things from a user-friendly way. I'd say something more like "updates were made to make sure the footer month and year autopopulate where possible". I'll use that, just wanted to let you know.

PR #1537, opened 2015-12-10
These updates make it testable

courtrightd:
Moved the 2nd dialog into the loop so it is recreated each time with the new set of properties, fixed some variable issues with UNEA.

It should be at least useable to test other panels on monday.


theVKC:
Beautiful! This copies much of what I worked on last night, but yours is more complete.

I was also working on dynamically finding the row for FFY 2014 (so, Oct 2013), as we found in testing that sometimes it's on row 7 and other times it's on row 8. Basically EMSearch aught to do it. I hadn't gotten there yet, if you want to go to there. :)

@IlseFerris @C-Love

IlseFerris:
I won't be able to take a look again until Monday. @courtrightd or @C-Love go for it if you have the time.

PR #1542, opened 2015-12-14
Create HOUSING GRANT FIATER.vbs

StormageddonDLOA:
This new script will do the FIATing for the worker when a MFIP case is receiving a subsidy but is still eligible for the Housing Grant. Finishes issue #1318.


Script Files/ACTIONS/HOUSING GRANT FIATER.vbs
```diff
@@ -0,0 +1,159 @@
+'Created by Tim DeLong from Stearns County.
+
+'STATS GATHERING----------------------------------------------------------------------------------------------------
+name_of_script = "ACTIONS - HOUSING GRANT FIAT.vbs"
```
theVKC: @StormageddonDLOA: this says "FIAT" but the script name is "FIATER". Can we make this "FIATER" please?

Script Files/ACTIONS/HOUSING GRANT FIATER.vbs
```diff
@@ -0,0 +1,159 @@
...
+start_time = timer
+
+'LOADING FUNCTIONS LIBRARY FROM GITHUB REPOSITORY===========================================================================
```
theVKC: @StormageddonDLOA: This is the old FuncLib block. Can we update with the newest one?

theVKC:
@StormageddonDLOA: you also need to name the file "ACTIONS - XXXX". You're missing the "ACTIONS" part.

PR #1543, opened 2015-12-14
Removed call center information from Client contact script for Henn. …

IlseFerris:
…Co. only

removed Call center information for HC users as this is causing confusion.

BLIP: added custom dialog box for Hennepin County users that does not include call center information.


theVKC:
Closing because I want to workshop a solution that isn't Hennepin specific.

Issue #1375, opened 2015-11-10
Remove stray "New!!!" notices from MAIN MENU scripts


theVKC:
Easy win. We should leave the ones from October forward. Just need to look through each. Anyone want it?

theVKC:
I'm removing all the ones from October and back. By the time the release goes out, November will almost be a distant memory.

theVKC:
Can we remove them from November forward? So only scripts that are new as of November or December would be included in this thing.

StormageddonDLOA:
I believe they should all be updated now. Please let me know if I missed any.

PR #1544, opened 2015-12-14
Update ACTIONS - MAIN MENU.vbs

StormageddonDLOA:
Adds a button for the HOUSING GRANT FIATER and cleans up the "new for" info on scripts that are no longer considered new. Issue #1375.


theVKC:
@StormageddonDLOA, the funclib block is wrong on this one, too. Can you use the updated one in the repo?

PR #1552, opened 2015-12-15
Update and rename NOTES - LOBBY NO SHOW.vbs to NOTES - INTERVIEW NO S…

### LINKED ISSUE #1260: Interview No Show.. New script or Enhancement?
I have heard from a lot of my co-workers here at Stearns that they would like a case note script for Interview No Show/ No answer when the client does not show or does not answer the phone for a scheduled interview. 

I realize there is a NOMI Script, however this only applies to SNAP cases. 
We also have the Lobby No Show script, however that indicates the worker paged the client. We don't page our clients here at Stearns. Here the client checks in at the front desk and then the interviewer is notified. 

I'm trying to figure out if maybe we should be looking at an enhancement to Lobby No Show or do a separate script. These are some of the changes I we would want to look at if we did an enhancement: 

- The Lobby No show could be changed to include scheduled appts (not just sameday interviews). Have the "Client was paged in the lobby" and 1st and 2nd page times be seperate and non-mandatory fields so other counties who don't use this system can still use the script. Instead we would just need a field for "Interview time" 

- Maybe the header could say something more generic like " No show for interview" or "Attempted phone interview- No answer" depending on what the worker checks instead of "Attempted to page client in Lobby" ?

- This script already has a checkbox to send the Nomi which is fantastic for new applications. It would also be cool if it did the Nomi script for FS recerts too. This could eliminate the need for the separate NOMI sent action script. 

- The script could also include CAF information like CAF Date, what programs applying for, or if it was a recert. This could be helpful if client calls back to reschedule. 

@C-Love  @theVKC  Any thoughts on this?

MelissaFox-Stearns:
…HOW.vbs
Resolves #1260 
This script is for both Same Day Interviews and Scheduled Interviews. If Same Day Interview is selected, the script will run what was previously known as Lobby No Show script which creates a case note when a client no shows to a lobby interview. If Scheduled Interview is selected, the script will run a dialog to gather application and interview information to enter a case note. 
Script has also been updated to incorporate the MEMO- NOMI script and gives the option to update PND2 to client delay within that script.


Script Files/NOTES/NOTES - INTERVIEW NO SHOW.vbs
```diff
@@ -0,0 +1,439 @@
+'STATS GATHERING----------------------------------------------------------------------------------------------------
+name_of_script = "NOTES - LOBBY NO SHOW.vbs"
```
theVKC: @MelissaFox-Stearns, can you update this too, to reflect the new script name?

Script Files/NOTES/NOTES - INTERVIEW NO SHOW.vbs
```diff
@@ -0,0 +1,439 @@
...
+start_time = timer
+
+'LOADING FUNCTIONS LIBRARY FROM GITHUB REPOSITORY===========================================================================
+IF IsEmpty(FuncLib_URL) = TRUE THEN	'Shouldn't load FuncLib if it already loaded once
+	IF run_locally = FALSE or run_locally = "" THEN		'If the scripts are set to run locally, it skips this and uses an FSO below.
+		IF default_directory = "C:\DHS-MAXIS-Scripts\Script Files\" THEN			'If the default_directory is C:\DHS-MAXIS-Scripts\Script Files, you're probably a scriptwriter and should use the master branch.
```
theVKC: @MelissaFox-Stearns: this is the old FuncLib block. Can you update it with the new one, which is located in the main directory of this repository?

Script Files/NOTES/NOTES - INTERVIEW NO SHOW.vbs
```diff
@@ -0,0 +1,439 @@
...
+EMReadScreen cash2_pend, 4, 7, 74
+EMReadScreen emer_pend, 4, 8, 74
+EMReadScreen grh_pend, 4, 9, 74
+EMReadScreen fs_pend, 4, 10, 74
+EMReadScreen hc_pend, 4, 12, 74
+
+'Assigns a value so the programs pending will show up in check boxes
+IF cash1_pend = "PEND" THEN
+	cash1_pend = 1 
+	Else
```
theVKC: @MelissaFox-Stearns: These Else statements should typically not be indented like this. Preferred format:

```
IF X = Y THEN
     do_some_stuff
ELSE
     do_some_other_stuff
END IF
```

Script Files/NOTES/NOTES - INTERVIEW NO SHOW.vbs
```diff
@@ -0,0 +1,439 @@
...
+	Else 
+	hc_pend = 0
+End if
+
+
+
+'Display's the Dialog Box to imput variable information - includes safeguards for mandatory fields
+If same_day_interview = vbYes THEN
+	Do					
+		Do
```
theVKC: @MelissaFox-Stearns: you have two different versions of error proofing here. Can you make sure you use the ` `err_msg = ""...` logic for both sections?

theVKC:
@MelissaFox-Stearns, I added some comments. Overall minor updates needed. Great job otherwise!

MelissaFox-Stearns:
Thanks V- 
I am working on the updates. I grabbed a lot of stuff from older scripts which I didn't change when I incorporated them into my script so I will make sure I do that. 
Thanks for the feedback!

theVKC:
I just quick skimmed it, and yes, it looks better. Because you updated in the same branch your pull request was from, it has automatically updated the pull request, so no additional GitHub-stuff should be necessary. :)

I'll evaluate it fully a bit later. Getting caught up on another project today. :)

theVKC:
@MelissaFox-Stearns: can we update the MAIN MENU script to point to this newly retitled script? It should also reflect the new title as well. @C-Love may want to take note of that for Ramsey users. :)

PR #1608, opened 2016-01-07
added FMED and COEX expenses

IlseFerris:
removed DCEX, and corrected some spelling errors.


courtrightd:
@IlseFerris - line 312 needs the <> replaced with = , currently giving a type mismatch error.  (I thought it would be easier for you to just add a commit rather than me fork your pending pull request.)

IlseFerris:
Done and done. Thanks for catching that. This is why pulling code at 5 am isn't the best idea :)

Issue #1365, opened 2015-11-09
ENHANCEMENT REQUEST -- ACTIONS - TRANSFER CASE -- new worker MEMO?


RobertFewins-Kalb:
A user in SIR has asked about adding the option to add a new worker MEMO to the in-county case transfer script. This was deliberately left off when the script was originally developed as it would be an annoyance to the worker to have to field calls about why the case was transferred.

Thoughts?

My vote is that if it is added, it should not be defaulted to checked.

IlseFerris:
I don't love this as an option, but if this is something that counties could use, then ok with me. It should most definitely not be defaulted to checked.

theVKC:
I think we _could_ do this, but I would want really, really clear explanations as to what it does within the script (I'm thinking a Yes/No MsgBox with the text on it).

Even so, I'd call for input on this. I could see it being annoying for the receiving agency.

PR #1634, opened 2016-01-11
Update ACTIONS - TRANSFER CASE.vbs

Y-Lor-Y:
Updated an option "within County" transfer section to include option that allow worker check off if they want to send a spec/memo to client to inform them of a new worker with the following text:  " I am your new worker.  If you have any question, please contact me at the phone number below."
Also included that if the option above is checked, it would be included in the case note that a spec/memo is sent to client.

Resolving issue #1365

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - TRANSFER CASE.vbs
+++ b/Script Files/ACTIONS/ACTIONS - TRANSFER CASE.vbs
@@ -109,8 +109,8 @@ BeginDialog out_of_county_dlg, 0, 0, 236, 385, "Case Transfer"
 EndDialog
 
 
-BeginDialog within_county_dlg, 0, 0, 216, 240, "Case Transfer"
-  EditBox 65, 10, 50, 15, case_number
+BeginDialog within_county_dlg, 0, 0, 211, 270, "Case Transfer"
+  EditBox 70, 10, 50, 15, case_number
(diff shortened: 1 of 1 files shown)


Script Files/ACTIONS/ACTIONS - TRANSFER CASE.vbs
```diff
@@ -125,20 +125,22 @@ BeginDialog within_county_dlg, 0, 0, 216, 240, "Case Transfer"
   CheckBox 170, 125, 40, 10, "EMER", EMER_pend_check
   DropListBox 100, 140, 65, 10, "Select one..."+chr(9)+"N/A"+chr(9)+"Yes"+chr(9)+"No", preg_y_n
   EditBox 85, 160, 120, 15, Transfer_reason
-  EditBox 80, 180, 125, 15, Action_to_be_taken
-  EditBox 80, 200, 125, 15, worker_signature
+  EditBox 85, 180, 120, 15, Action_to_be_taken
+  CheckBox 10, 200, 195, 10, "Check to send out a SPEC/MEMO to client of new worker.", spec_memo_within 'add this as option for worker to use for w/in county xfer.
```
theVKC: This should probably have a `_check` at the end of the variable (`spec_memo_check`). Not sure what `_within` refers to.

PR #1665, opened 2016-01-14
BULK CASE TRANSFER

### LINKED ISSUE #956: BULK case transfer in MAXIS
I was approached by a Hennepin County supervisor requesting a BULK case transfer script that can do the following:
1.  Allow the user to select certain criteria within the cases (post 60 month MFIP cases, immigration codes, sanction cases, etc)
2. Allow the user to select one or multiple users to transfer the cases to, and vice versa. 
3. Create an Excel report of the cases transferred.

C-Love:
Resolves #956 
New script!

Blip - Searches caseload(s) by selected parameters. Will then transfer a specified number of those cases to another worker. - End Blip


theVKC:
FYI, folks in @MN-Script-Team/5k-club: thinking we should all test this a bit. :) Perhaps talk to some of our sups too about testing. Button about to go live in master!

PR #1668, opened 2016-01-15
Adding 4 month retaining

### LINKED ISSUE #364: Inac scrubber 4 mo reinstate
So with this new 4 month reinstate for MAGI clients our supervisors here have asked us to redo the inac scrubber to avoid these cases so they can sit in the worker's case load for the 4 month reinstate period. 

We wanted to know what other agencies are doing or if anyone else was looking at solutions for this.

CDPotter:
BLIP: Split script into two paths, one that will retain MAGI cases closed for no/incomplete renewals (along with all previous qualities EX open in MMIS, unresolved dails messages, denials.) in current worker number. The other that will transfer the case to closed(after checking against previous qualities) even if it is a MAGI case closed for no/incomplete renewal. The script works in conjunction with the global variables set by each county agency so agencies can decide what works best for their workflow.

Resolves #364 (Did you want me to add variable to GLOBVAR or will that be done after the install?


theVKC:
This change is 1092 lines long. It's not showing as something displayable on GitHub, so I have literally no idea what you're pulling. Can you please split it into 3 commits so that I can see it all? Trying to avoid problems.

Also: yes, can you add the globvar variable and set it to false?

CDPotter:
So divide it by 3 and commit 3 separate times, add the variable to Global variables then pull that branch?

theVKC:
Sounds great. That way I can evaluate each commit before pulling it in (I reeeeally don't want to have to clone your branch and work offline without a diff)

PR #1669, opened 2016-01-15
Adding 4 month retaining to inac scrubber

### LINKED ISSUE #364: Inac scrubber 4 mo reinstate

CDPotter:
BLIP: Split script into two paths, one that will retain MAGI cases closed for no/incomplete renewals (along with all previous qualities EX open in MMIS, unresolved dails messages, denials.) in current worker number. The other that will transfer the case to closed(after checking against previous qualities) even if it is a MAGI case closed for no/incomplete renewal. The script works in conjunction with the global variables set by each county agency so agencies can decide what works best for their workflow.

Resolves #364


theVKC:
Looks good. Thank you!

Issue #1702, opened 2016-01-20
BULK DAIL report has weird spacing issue


theVKC:
It's occasionally writing large amounts of spaces in the Excel sheet. On it.

PR #1703, opened 2016-01-20
Resolved #1116 and #1702

### LINKED ISSUE #1116: BULK DAIL report option to use on multiple users (enhancement)
Request rec'd from Hennepin County supervisor who would like to be able to pull information from several user's DAIL messages.

theVKC:
Yay! See commit.


theVKC:
My commits got messy, so here's the comment:

Let's see...

* Added more space to write in the dialog (for more x1 numbers
* Added extra MAXIS checks before and after the dialog
* Splits x1 numbers into an array
* Array handling added to do...loop
* Adjusted time savings to be based on line copy/pasting (@GayS, in case
you're keeping a list, it's 25 seconds/item now)
* Added x1 number to Excel list
* Added basic usage stats to the end (want to show people how many
messages/time saved)
* Added logic to make sure we don't move ahead on the Excel list without
having data in the Excel list (prevents weird spacing where excel_row
was increasing without cause)

Blip: the script now allows users to enter more-than-one x1 number.
Also, enhancements were made to prevent blank rows from being added to
Excel, and basic statistics on script runtime are now included in the
final Excel document.

PR #1719, opened 2016-01-21
Fixed GA bene month bug, and HCRE

CDPotter:
GA bene month was running through dateadd and turning something like 12/15 into 12/15/16

HCRE function is completely useless. If HC is added to TYPE MAXIS automatically navs to HCRE and enters the app info. We have no need to train on a retro case (and approval cannot handle that either) I removed the write to HCRE function. At this time I would suggest just striking through the HCRE portion on the excel spreadsheet and telling people not to fill it out rather than deleting the rows and reprogramming all of locations. The function also contained the now deleted MAXIS_dater function. Should we leave this in funclib for the time being or just delete the write to HCRE function?

### DIFF
--- a/Project Krabappel/BULK - project Krabappel.vbs
+++ b/Project Krabappel/BULK - project Krabappel.vbs
@@ -568,7 +568,7 @@ For each case_number in case_number_array
 		call create_MAXIS_friendly_date(APPL_date, 0, 10, 44)
 		call create_MAXIS_friendly_date(APPL_date, 0, 10, 55)
 	End if
-	If HC_application = True then call create_MAXIS_friendly_date(APPL_date, 0, 12, 33)		'No interview or elig begin dt for HC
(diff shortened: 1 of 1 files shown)


theVKC:
Wondering: do you think we might want to include it (but only insofar as someone would maybe want to APPL a retro case for training purposes), but include language in Excel that says "hey look, you won't be able to approve this if anything is in here?"

I could foresee a situation where someone wants to update HCRE with a retro date, but leave it in PND2 for training purposes.

CDPotter:
The way Robert and I talked through it was you can always update HCRE on your at any time. So for this hypothetical training situation "Case is pending and client calls to request retro" so the worker can get practice updating HCRE and adding the retro month FIATs. 

(Things get complicated when since HCRE is case based BUT also person based at the same time.)

theVKC:
I guess I'd rather not do the strikethrough approach. In my opinion, it looks sloppy. We can either remove the panels from the spreadsheet or ensure the functionality works.

CDPotter:
I'll get deleting then.

PR #1765, opened 2016-01-28
Fixing Bug that transfers too many cases

### LINKED ISSUE #1764: BUG - Bulk Case Transfer is not stopping at the specified number of cases
Basically the script will just keep on transferring until it gets through all of the cases, completely ignoring the number put entered in the second dialog. It doesn't happen every time. 

I have the fix (convert a string to a number so math will happen).

C-Love:
Resolves #1764


theVKC:
Can I get a blip for this, @C-Love?

C-Love:
Blip - 'Fixes an issue where the script would transfer more cases than was requested'

theVKC:
Can you also send this through to release?

PR #1869, opened 2016-02-16
Resolves #1867 Adding correct stats handling BULK CN from list

### LINKED ISSUE #1867: BULK MEMO FROM LIST, stats counter only counting as one
@CDPotter, I was looking over your script to update the script list w/ time savings, and it looks like the stats counter is only counting the usage of the script as one instance, instead of it counting each instance within the script. Can you take a look at this one?

CDPotter:
BLIP: Corrected time savings counted to correctly reflect time savings from script.

### DIFF
--- a/Script Files/BULK/BULK - CASE NOTE FROM LIST.vbs
+++ b/Script Files/BULK/BULK - CASE NOTE FROM LIST.vbs
@@ -47,7 +47,7 @@ END IF
 'Required for statistical purposes==========================================================================================
 STATS_counter = 1                          'sets the stats counter at one
 STATS_manualtime = 180                               'manual run time in seconds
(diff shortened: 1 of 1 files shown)


Script Files/BULK/BULK - CASE NOTE FROM LIST.vbs
```diff
@@ -325,4 +326,5 @@ IF privileged_array <> "" THEN
 	MsgBox "The script could not generate a memo for the following cases..." & vbCr & privileged_array
 END IF
 
+STATS_counter = STATS_county - 1  'subtracts one from the stats (since 1 was the count, -1 so it's accurate)
```
theVKC: @CDPotter: same issue on this one. Should read `STATS_counter`.

PR #1870, opened 2016-02-16
Resolves #1867 Adding correct status handling BULK MEMO from list

### LINKED ISSUE #1867: BULK MEMO FROM LIST, stats counter only counting as one

CDPotter:
BLIP: Corrected time savings counted to correctly reflect time savings from script.

### DIFF
--- a/Script Files/BULK/BULK - MEMO FROM LIST.vbs
+++ b/Script Files/BULK/BULK - MEMO FROM LIST.vbs
@@ -47,7 +47,7 @@ END IF
 'Required for statistical purposes==========================================================================================
 STATS_counter = 1                          'sets the stats counter at one
 STATS_manualtime = 180                              'manual run time in seconds
-STATS_denomination = "I"       'I is for each Item
+STATS_denomination = "C"       'C is for each case
 'END OF stats block==============================================================================================
 
 'Dialogs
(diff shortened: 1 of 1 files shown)


theVKC:
@CDPotter: can you tell me why this is three separate pull requests? I'd do separate commits, one pull request (doing three pull requests makes things more challenging on my end to catalog and process).

I won't make you redo it. Just curious is all.

CDPotter:
Figured since they were all separate scripts I'd do them separately in case any issues arose and we needed to backtrack.

theVKC:
The way we'd normally do that is separate commits, but one pull request. I can roll back separate commits as easily as pull requests... separate pull requests just means I have more to track on release notes and internal tracking. Anything we can do to save me some time is very much appreciated. :P

Script Files/BULK/BULK - MEMO FROM LIST.vbs
```diff
@@ -334,4 +335,5 @@ IF privileged_array <> "" THEN
 	MsgBox "The script could not generate a memo for the following cases..." & vbCr & privileged_array
 END IF
 
+STATS_counter = STATS_county - 1  'subtracts one from the stats (since 1 was the count, -1 so it's accurate)
```
theVKC: Same goes here, @CDPotter: `STATS_county` should be `STATS_counter`.

Issue #710, opened 2015-06-15
project mic drop


IlseFerris:
A plan to build a comprehensive eligibility review interview scheduler for SNAP and MFIP.

Task list:
- [x] Asks for dates on dialog (save the first 8 days of the month)
- [x] Goes into each case on REVS and logs the ones with a review requiring an interview
- [x] Adds day/time of interview based on earlier dialog
- [x] Shows an Excel list of all of the cases
- [x] Case notes each case
- [x] Sends Appointment letter for each case
- [x] Includes developer mode (#713)
- [x] Includes link to instructions page (#1700)
- [x] Includes multi-user functionality (can run on multiple caseloads) (#1084) - Assigned: @theVKC
- [x] The user can cap the reviews scheduled per worker (#1083) - Assigned: @theVKC
- [x] A `create_dynamic_calendar_dialog` function which can output selected dates as an array (will be moved to FuncLib when completed) (#1711) - Assigned: @theVKC
- [x] A 'create_outlook_appointment` function which can optionally update your Outlook calendar with all scheduled dates. - Assigned: @theVKC
- [x] Statistics gathering handling - Assigned: @IlseFerris
- [x] A separate DAIL scrubber script to process the missed interviews (#711)
- [x] Remove Veronica's elitist functionality - Assigned: @CDPotter
- [x] Standardize date format for the TIKL (MM/DD/YYYY) - Assigned: @theVKC
- [x] Add to bulk MAIN MENU before pulling into master - Assigned: @IlseFerris
- [x] Research why on-the-hour appointment scheduling in Outlook does not schedule out for the time period selected (e.g.: 30 minute appt only scheduled from 800-800) 
- [x] Date on notice that paperwork needs to be submitted to the agency by has an excess "/" mark

theVKC:
@IlseFerris, I'm taking you off of this as the assignee. This is clearly being worked on by a team, and I don't need it mucking up your task list. :)

theVKC:
Task list update allows me to close #1083, #1084, and #1711.

theVKC:
FYI all: I'm targeting this for the March release. I will be formally closing the Mic Drop milestone today.

theVKC:
@IlseFerris: see #1889 about a planned upgrade to the BULK Main Menu... Perhaps that step should come before adding this script?

theVKC:
[...]

PR #1890, opened 2016-02-18
small fixes

CDPotter:
1: max_reviews_per_worker is getting ABS'd which causes type mismatch error: Made conditional based on if it's entered
2: the text for the maximum interviews is getting cut off: Fixed


theVKC:
This addresses #710.

PR #1897, opened 2016-02-19
Resolved #1891, Pay stubs rec'd pic case note bug

### LINKED ISSUE #1891: Pay stubs rec'd- case noting the PIC all wonky
Has anyone gotten reports of the SNAP PIC is not case noting correctly?
image

This seems to be happening in both branches. 
This is from release branch: 
image

IlseFerris:
BLIP: Script has been enhanced to case note the PIC correctly if more then 4 pay stubs are being used in the script.


theVKC:
I'm going to hold off on this solution and target it for the March release. I see some stuff in here that I would want properly tested by users, @IlseFerris. The MsgBox parts have me q

PR #1933, opened 2016-02-25
Resolved #1821- Check for expiration

### LINKED ISSUE #1821: Dail - New Hire- Have script check jobs panel carries through to next footer month
When New Hire script adds a jobs panel in, often times it does not carry through to the next month. 
It seems it would not be possible to have the script go into the next footer month to add or check for the jobs panel because it could cause a duplicate jobs panel to be entered. 
What I am proposing is to have the New Hire script check for the "This data will expire on ...." when Jobs panel is entered and then alert the worker that it may not have carried over. 

Any thoughts?

MelissaFox-Stearns:
Added feature to New Hire Dail Scrubber. When the script enters a Jobs panel, it checks for the "This Information will expire ..." which indicates that the JOBS panel will not carry forward to the next footer month. When this happens the script will now give you a Msgbox to alert the worker to check the next footer month.

### DIFF
--- a/Script Files/DAIL/DAIL - NEW HIRE.vbs
+++ b/Script Files/DAIL/DAIL - NEW HIRE.vbs
@@ -219,7 +219,10 @@ If create_JOBS_checkbox = checked then
 		transmit
 	End if
 	transmit						'Transmits to submit the panel
+	EMReadScreen expired_check, 6, 24, 17 'Checks to see if the jobs panel will carry over by looking for the "This information will expire" at the bottom of the page
+		If expired_check = "EXPIRE" THEN Msgbox "Check next footer month to make sure the JOBS panel carried over"
 End if
+
 
 'Navigates back to DAIL
 Do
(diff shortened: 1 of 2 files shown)


Script Files/DAIL/DAIL - NEW HIRE.vbs
```diff
@@ -228,6 +231,7 @@ Do
 	PF3
 Loop until DAIL_check = "DAIL"
 
+msgbox "Do you want to case note?"
```
theVKC: @MelissaFox-Stearns: what is this MsgBox for?

theVKC:
@MelissaFox-Stearns: please see my comment. Thank you?

MelissaFox-Stearns:
Hey!  Did I put that in there? I saw that when I was testing but I don't remember adding that message box so I thought it must have been there before.

theVKC:
Looks great, thank you!

PR #1942, opened 2016-02-29
RESOLVES #1941 fixing maxis footer month

### LINKED ISSUE #1941: BUG CAF not reading WREG months
We changed the variable footer_month to MAXIS_footer_month in the CAF but the autofill functions all still use footer_month. Should we just change CAF back to footer_month? PRISM doesn't use footer_months so we don't need to differentiate.

CDPotter:
BLIP: Script was updated to change variable name for footer month. this caused script to be unable to use functions that also didn't have this standard variable name updated. One the one reported function the script was unable to accurately reading the counted months off of WREG because footer_month was considered blank.

 Script can now work with these functions to the fullest again.


theVKC:
When you say "unable to use functions", can you expand on that, @CDPotter? Does the script error out entirely, or does it continue without noting anything?

CDPotter:
Updated.

PR #1969, opened 2016-03-07
Resolves #1967 Updating Citizenship DAIL Scrubber

### LINKED ISSUE #1967: Enhancement -- DAIL >> Citizenship Verified
A case reviewer in Anoka has noticed a common error on cases involving the citizenship verified DAIL scrubber. The user has been allowing the script to put "OT" as the verif code for all persons in the household when we have a specific verif on file for the client's citizenship. The assumption is that the user did not double check who needed to be updated and had the script update for all peeps. After some conversation, the idea was forwarded to require the user to indicate which household members to update rather than indicating which household members to not update.

In the custom function HH_member_custom_dialog_cit_id_ver, we would be defaulting the value for all_clients_array(x, 1) to 0 instead of 1, requiring an affirmative decision on the part of the user.

RobertFewins-Kalb:
BLIP: Updating the Citizenship/ID DAIL scrubber to require that the user select the specific household members to update, instead of the household members to not update. The script will not allow you to continue until you have selected at least one household member, and that (those) member(s) should be listed on DAIL/DAIL.

### DIFF
--- a/Script Files/DAIL/DAIL - CITIZENSHIP VERIFIED.vbs
+++ b/Script Files/DAIL/DAIL - CITIZENSHIP VERIFIED.vbs
@@ -79,7 +79,7 @@ Function HH_member_custom_dialog_cit_id_ver(HH_member_array)
 	FOR x = 0 to total_clients				'using a dummy array to build in the autofilled check boxes into the array used for the dialog.
 		Interim_array = split(client_array, "|")
 		all_clients_array(x, 0) = Interim_array(x)
-		all_clients_array(x, 1) = 1
+		all_clients_array(x, 1) = 0			'Defaulting to update none persons so the user has to update them thar persons
 	NEXT
 
 	BEGINDIALOG HH_memb_dialog, 0, 0, 191, (35 + (total_clients * 15)), "HH Member Dialog"   'Creates the dynamic dialog. The height will change based on the number of clients it finds.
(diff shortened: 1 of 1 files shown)


theVKC:
Anyone think there's value in modifying the primary function to allow parameters to determine who to check? I could foresee an array with info from eats, a blank array, or a filled array. Thoughts?

In any case, I like this mod. I'll incorporate it when I'm back at my desk. Thanks @RobertFewins-Kalb!

CDPotter:
Might work, the way the dail is displayed (I believe) it's "Applicant name (affiliated name)" similar to school dails. So it would have to read that and compare it to what's written on MEMB to get a MEMB #.

RobertFewins-Kalb:
@theVKC, are you thinking something like...

create_HH_member_array(major_program)
  code
  code
  some more code
  IF major_program = "HC" THEN 
    code to build group from HCRE
  ELSEIF major_program = "SNAP" THEN 
    code to build group from EATS
  ELSEIF major_program = "EMER" THEN 
    code to build group from MEMB
...


... because I could see application with consolidating dynamically scripts based on the action in question.

Scenario: User needs to case note intake...
- Initial case number and programs dialog, indicate intake...

FOR EACH major_program IN programs_array
  CALL create_HH_member_array(major_program)
  code for getting down to business...
NEXT

theVKC:
Yes! And then the dialog would merely display the contents of this array (allowing customization).

Any interest, @RobertFewins-Kalb? I'd be putting it on FuncLib as an issue instead of the main repo. I could see this being really helpful for lots of autofill scripts.

PR #1971, opened 2016-03-09
Resolves #1970 script reads entire date/time now

### LINKED ISSUE #1970: BUG: DAIL - SEND NOMI not entering dates right
I was running the dail scrubber on my revs scrubber messages (granted I don't remember if we used a local version of revs scrubber) and the client didn't turn in CAF so I selected don't sent nomi option. It entered this as the case note.

2016-03-09 10_03_20-s1 - anoka desktop - anoka scripted desktop - bluezone mainframe display

Here is the dail:

dail

CDPotter:
BLIP: Changed way script reads TIKL by just reading entire date and time in one line and making it one variable.

Figured that reading the entire rest of the sentence would make sense. I also accounted for an extra 3 digits for MM DD and HH

### DIFF
--- a/Script Files/DAIL/DAIL - SEND NOMI.vbs
+++ b/Script Files/DAIL/DAIL - SEND NOMI.vbs
@@ -83,9 +83,9 @@ EMConnect ""
 EMSendKey "x"
 transmit
 
-'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW AT..." This is the part that is static in the DAIL message
-EMReadScreen interview_date, 10, 9, 42
-EMReadScreen interview_time, 8, 9, 56
+'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW APPT AT..." This is the part that is static in the DAIL message
+EMReadScreen interview_date_time, 19, 9, 46
+'EMReadScreen interview_time, 8, 9, 56
 row  = 1
 col = 1
 EMSearch "Case Number: ", row, col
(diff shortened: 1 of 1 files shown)


Script Files/DAIL/DAIL - SEND NOMI.vbs
```diff
@@ -83,9 +83,9 @@ EMConnect ""
 EMSendKey "x"
 transmit
 
-'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW AT..." This is the part that is static in the DAIL message
-EMReadScreen interview_date, 10, 9, 42
-EMReadScreen interview_time, 8, 9, 56
+'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW APPT AT..." This is the part that is static in the DAIL message
+EMReadScreen interview_date_time, 19, 9, 46
+'EMReadScreen interview_time, 8, 9, 56
```
theVKC: Can we clear this out?

Script Files/DAIL/DAIL - SEND NOMI.vbs
```diff
@@ -83,9 +83,9 @@ EMConnect ""
 EMSendKey "x"
 transmit
 
-'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW AT..." This is the part that is static in the DAIL message
-EMReadScreen interview_date, 10, 9, 42
-EMReadScreen interview_time, 8, 9, 56
+'Reading date and time of recertification appointment from the TIKL--DAIL message that should be read is: "~*~*~CLIENT HAD RECERT INTERVIEW APPT AT..." This is the part that is static in the DAIL message
+EMReadScreen interview_date_time, 19, 9, 46
```
theVKC: I like this solution! But, is it possible to trip it up with a REALLY long date and time? Like: "10/10/2016 at 12:45 PM"? I'd bet that's the char maximum... does the TIKL overflow in these instances?

theVKC:
See comments, @CDPotter. Thanks!

Script Files/DAIL/DAIL - SEND NOMI.vbs
CDPotter: I made it 19 to account for those. The TIKL is exactly long enough to handle it that full date.

2016-03-10 08_55_41-s1 - anoka desktop - anoka scripted desktop - bluezone mainframe display

"MM/DD/YYYY HH:MM PM)

CDPotter:
Fixed em.

PR #1976, opened 2016-03-10
Resolves #1975 -- Updating income limits

### LINKED ISSUE #1975: Bug -- LEP Sponsor Income income standards have changed
An Anoka LEP worker has notified me that the income standards have changed and the script is not up to date. See: CM 19.06

What are peeps' thoughts on adding constants like this to the FuncLib? That way, during COLA, we are able to catch all the changing income standards at once instead of updating scripts piecemeal.

RobertFewins-Kalb:
BLIP: Updating the constants for the 2016 sponsor income limits. See: CM0019.06.


theVKC:
Thanks for the CM reference in the blip, @RobertFewins-Kalb!

PR #1986, opened 2016-03-14
Resolved #1980, resolved #1978: removed CS disregard FIATER and WCOM 

### LINKED ISSUE #1978: Remove CS disregard fiater from existence
MAXIS has now been updated to handle this. It needs to be removed since running it over the cases since 10/15 will cause the disregard to be duplicated.

### LINKED ISSUE #1980: Remove CS disregard WCOM from scripts
MAXIS also updated the notice so this is no longer needed.

IlseFerris:
BLIP: Retired the FIATER and WCOM scripts as MAXIS has been updated to handle processing of these cases. Users will no longer need to manually FIAT eligibility, or added notes to worker comments.


theVKC:
Main menu also needs updating. Will pull that in first.

PR #2003, opened 2016-03-14
Resolved #1993: new case note for returned mail

### LINKED ISSUE #1993: New script idea - detailed returned mail scipt
This wouldn't be a bulk script but would be a case specific script that would go into much greater case note detail for the counties that like that info.  I'll take it.

StormageddonDLOA:
adds a case note that details the ADDR rec'd from, new ADDR if one is known, county of new ADDR if applicable, if returned mail was resent to the client, if forms were sent to the client, if the client has a MNsure case, the MNsure number if applicable, if the MNsure case has the correct ADDR, misc notes/actions taken. Automatically capitalizes the address (old and new) and the county when case noted.


Script Files/NOTES/NOTES - RETURNED MAIL.vbs
```diff
@@ -0,0 +1,239 @@
+'STATS GATHERING----------------------------------------------------------------------------------------------------
+name_of_script = "NOTES - RETURNED MAIL RECEIVED.vbs"
```
theVKC: This should match the name of the file, @StormageddonDLOA.

PR #2028, opened 2016-03-18
Tackling MF = MS issue

CDPotter:
changed logic to prevent issues where MS and ACTV renewals were being assessed as MF and ACTV thus not deleting them if there was a SNAP CSR or renewal in 12 more months.


theVKC:
@IlseFerris: you might want to investigate this.

CDPotter:
Also @C-Love if Ramsey has started testing this.

PR #2033, opened 2016-03-21
Resolves #2032 bug in Notices - NOMI

### LINKED ISSUE #2032: bug in NOTICES - NOMI missing END IF statement
This renders it unusable, looks like it happened with last change.  Master branch only, fix inbound.

courtrightd:
Putting back the missing END IF for the x127 specific section.


theVKC:
Ack!

PR #2036, opened 2016-03-23
Resolves #2035 Updated Denial reason

### LINKED ISSUE #2035: NOTES- Significant Change- Denial reason update
One of the denial reasons for Significant change was "income didn't decrease by 50%". Well, in October they changed the policy to decrease $65 and 50%. 
I want to change to update the denial reason to be something like.... "Income didn't decrease enough"
so we don't have to update it every time they make a change to policy. :)

MelissaFox-Stearns:
Changed denial reason "Income didn't decline at least 50% in the benefit month" to "Income did not decrease enough to qualify" since the policy on Significant Change is different from when the script was originally written.


theVKC:
Awaiting @MelissaFox-Stearns's response on #2035 before deciding whether-or-not to pull this.

MelissaFox-Stearns:
I just changed the wording to fit policy. I'm not sure what it would be considered exactly.

PR #2043, opened 2016-03-28
Resolves #1964 Adding MSA training case

### LINKED ISSUE #1964: Training Case Creator: needs an MSA case
Spinning this off as a separate issue from #53, because it is unlikely this will be done by the March deadline I set.

@courtrightd has taken this on, and we'll time it for the April release. Yay!

CDPotter:
BLIP: Added MSA case including MSA/HC/SNAP

(NOTE: MSA/SNAP sometimes get approved back to back so quickly the initial month doesn't have MSA budgeted into SNAP. Issue will be created and fixed after freeze.)

No idea what those other commits are or why they are pulling through. I'm blaming github for windows.


theVKC:
One is the file from GitHub online. Looks like that caused a hitch. Can you delete the one with all the periods for me?

CDPotter:
Deleted.

PR #2067, opened 2016-03-31
Resolves #2065 Returning to selected JOBS panel

### LINKED ISSUE #2065: Bug -- Paystubs received is jumping ahead one employer.
I suspect the checking for password out modification is what is causing this.

Staff are reporting that the script is putting the information on the wrong panel. I think the transmit in check_for_password is doing it.

I'm looking into it

RobertFewins-Kalb:
Blip: The script has been programmed to check for MAXIS timing out, but that appears to be forcing through a transmit which jumps ahead 1 JOBS panel. The script we now re-write the desired JOBS panel number to undo that jump. (Now for release as well)

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - PAYSTUBS RECEIVED.vbs
+++ b/Script Files/ACTIONS/ACTIONS - PAYSTUBS RECEIVED.vbs
@@ -267,7 +267,12 @@ LOOP UNTIL ButtonPressed = -1 AND number_of_paystubs <> "" AND IsNumeric(number_
 
 DO
 	CALL create_paystubs_received_dialog(worker_signature, number_of_paystubs, paystubs_array, explanation_of_income, employer_name, document_datestamp, pay_frequency, JOBS_verif_code)
-
+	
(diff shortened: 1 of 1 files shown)


theVKC:
Two pull requests for Paystubs Received, @RobertFewins-Kalb? Was that an accident?

RobertFewins-Kalb:
Release and master. I found the issue in both branches

theVKC:
Both are master. Also, does this issue meet critical status?

RobertFewins-Kalb:
Poop. Rookie mistake. I think it meets critical status. If the client has multiple JOBS panels, and the user is trying to update any panel other than the last one, it does not work.

theVKC:
I'm guessing it isn't readily obvious it didn't work? As in the end user might think it worked properly?

CDPotter:
Yup, I assuming someone ignored a few warning signs in a budget and could edit job 2 instead of job 1 and have two higher paying jobs counted and budgeted.

theVKC:
OK, I can buy that as critical. @RobertFewins-Kalb, if you can pull into release as well, that'd be great.

PR #2068, opened 2016-03-31
Resolves #2060 - Adding Verification Still Needed

### LINKED ISSUE #2060: STILL NEEDED - MEMO/CN
When several verifications are received (AFTER 2919 has been sent), but some are outstanding this will CN what is left, and send a MEMO to the client letting them know what is still needed. Will be useful to bank counties when a worker picks up a case after another worker processed a pending application, incomplete renewal, incomplete HRF, etc... In the future it may be useful to connect it to the 'docs received" script. Will also include some information to help make sure that workers used the 2919 for the first request of any verification.

LaLarson:
BLIP: This script is meant to be used when several verifications have been received from the client but there is still outstanding information needed to determine eligibility. It will case note all outstanding information entered in to the script and then open a Word document, and list the information there as well to be sent out to the client; informing them that more information is still needed.


LaLarson:
Unmerged commits?? That doesn't sound good?? Can I just delete the branch? It looks like it wants me to delete the branch!

theVKC:
This one you can delete. It's the duplicate from using GitHub on the web.

theVKC:
Don't delete the "still needed" one though!

PR #2084, opened 2016-04-06
RESOLVES #2047 Reworking room&board logic

### LINKED ISSUE #2047: Shelter Expense Verif Recv'd Issue - Room and Board
A worker used this script and selected the "Check here if this is for Room and Board" and finished the rest of the fields. When she selected "OK" a pop up window opened informing her that she would need to enter in the information into the SHEL panel manually. The script then created 2 separate case notes. One case note had the correct title, how much the rent was ($0 in this case), and the worker's signature. The second case note had a weird title (the number 3 and some symbols), and then listed the new address that was reported.

CDPotter:
BLIP: Logic surrounding room & board was incorrect, fixed IF statement and added section to actually case note room & board notes.


theVKC:
I'll take your pull request, @CDPotter, but your blip could use some work. The average user doesn't know what an "if" statement means.

theVKC:
I'll create my own blip, but consider the audience in the future. :)

PR #2090, opened 2016-04-07
RESOLVES #2044 Adding note about MSA into SNAP budget

### LINKED ISSUE #2044: bug fix: Training case gen MSA/SNAP approval needs to slow down
MSA and SNAP get approved too quickly back to back causing the first month of SNAP to not have the very recently approved MSA budgeted into it. 

I'll take a crack at this.

CDPotter:
BLIP: Note added to warn users that on MSA/SNAP cases the MSA grant may not be budgeted for the initial month of SNAP.


theVKC:
Great job!

PR #2113, opened 2016-04-13
RESOLVES #2098 correcting verbiage/adding state_a_blank_case_note

### LINKED ISSUE #2098: Bug Dail scrubber REVS scrubber verbiage/inquiry handling
Notices two things today. 

One I accidentally ran it in inquiry and it got into an infinite loop trying to pf9 on case note. 

Two the verbiage on this seems contradictory to me. I had a client who didn't contact and didn't send in CAF so it seems according to the sentence send a nomi but according to "press no" statement click no?

nomi

CDPotter:
BLIP: verbiage on question about when to send NOMI was reworked to read better to end users. Corrected issue that caused infinite loop when trying to run in inquiry.

### DIFF
--- a/Script Files/DAIL/DAIL - SEND NOMI.vbs
+++ b/Script Files/DAIL/DAIL - SEND NOMI.vbs
@@ -107,7 +107,7 @@ If interview_confirm = TRUE then
 	script_end_procedure("Success! A NOMI is not required if the recertification interview is complete." & vbNewLine & "Please review the case for completion if necessary.")
 ELSE
 	'Msgbox asking the user to confirm if the client has sent a CAF or if no contact has been made by the client
-	recert_forms_confirm = MsgBox("A NOMI is needed when a SNAP recipient has not made contact with the agency about their recertification, AND the CAF has not been received." & vbNewLine & vbNewLine & "Press Yes to send the NOMI." & _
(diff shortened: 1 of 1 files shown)


Script Files/DAIL/DAIL - SEND NOMI.vbs
```diff
@@ -117,7 +117,7 @@ END IF
 If result_of_msgbox = FALSE then		'if false a case note will be made, but a NOMI will not be sent as this is not necessary. 
 	dialog worker_signature_dialog
 	If ButtonPressed_worker_signature_dialog = 0 then stopscript
-	PF9	'goes directly into edit mode
+	Call start_a_blank_case_note 'Navigates to a blank case note & writes the case note
```
theVKC: I don't think start_a_blank_case_note works from the DAIL, @CDPotter. It forces people out of the DAIL workflow and then they can't get back to their message without a bunch of navigation. Have you tested this? Does F3-ing out of the case note navigate back to DAIL, or SELF?

CDPotter:
No I didn't test that functionality of it. I was following the way it was written down below. The other section does the same thing. I could probably pull apart the start a blank case note and put the do loop in there manually?
nomi

theVKC:
I don't know. I don't think `start_a_blank_case_note` works as expected when navigating from the DAIL, same as `navigate_to_MAXIS_screen`. That's a bigger discussion, I think.

CDPotter:
Well in testing Navigate to MAXIS screen it should try to use the case_number and will bring you to DAILS for that specific case, just not for sure the exact DAIL you left off on. 

I'd argue that since this script doesn't need to get to other dails, or delete dails that this method for getting back to DAILS works for this case.

theVKC:
Sounds good. Thanks @CDPotter!

PR #2119, opened 2016-04-14
Resolves #2103 PA Verif FTI verbiage

### LINKED ISSUE #2103: PA VERIF - add some verbiage on the dialog warning users not to share FTI from SSA
We received information during the Maxis Mentor call 2/24/16 from Brian Shields stating we are not allowed to verify information regarding income from other sources such as SSI/RSDI, VA benefits, income form a job.  However, this script pulls that information.  Our understanding is that this is from an FTI rule.  Can you clarify if we:
1. Can include that information?
2. Are supposed to take that information off before sending the letter 
3. Can it be corrected so it doesn’t pull that information?
Thanks,
Bridget Blomer
Scott County

courtrightd:
Blip: Adds a reminder on the dialog that users should not share FTI information with other agencies by using this form.

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - PA VERIF REQUEST.vbs
+++ b/Script Files/ACTIONS/ACTIONS - PA VERIF REQUEST.vbs
@@ -71,11 +71,10 @@ BeginDialog case_number_dialog, 0, 0, 151, 70, "PA Verification Request"
 EndDialog
 
 
-BeginDialog PA_verif_dialog, 0, 0, 190, 250, "PA Verif Dialog"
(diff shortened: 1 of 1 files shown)


theVKC:
I don't know about "Do not share FTI on this form, including information from SSA such as SSI/RSDI amounts.", as it might lead folks to assume they can't share it with anyone, including clients or AREPs. Can we tweak the wording a bit?

courtrightd:
“Warning: Do not share FTI with outside agencies using this form, including information from SSA such as SSI/RSDI amounts.”
Does that work better?

theVKC:
Much better.

PR #2180, opened 2016-05-09
RESOLVES #2179 Adding safeguard for non matching new hire names

### LINKED ISSUE #2179: New Hire message (new hire name <> MAXIS name)
So a coworker and myself both had issues this month where we failed to notice that the newhire message was for someone with a different name than the MAXIS name of the person with that SSN. My case was a switch of the SSN number (likely the new employee's mistake) while someone else had completely different SSN numbers. 

I was thinking it might be good to compare the two strings and prompt the user if they don't match, allowing them to stop or continue (to account for missing middle initials and spacing differences)

If this is good to go I will get started on it.

CDPotter:
BLIP: When a client gets a new hire and someone else has used their SSN the names do not match. The script was not reviewing these inconsistencies. Now the script will pause if the names are not identical and give the user a chance to review the information and stop the script if necessary.


CDPotter:
waiting till after may release.

theVKC:
@CDPotter, I didn't know you had done this already when I left my comment.

Issue #1938, opened 2016-02-27
Assorted updates needed across all scripts


theVKC:
We should time this for a Monday hackathon, especially one in which there's a release (next DHS hackathon).

- [x] Update FuncLib block to remove the term "Alpha Users" and shorten the MsgBox (IT folks don't need to be involved, errors seldom come up and I am not hearing of HTTP errors)
- [x] Update all instances of `case_number` to read `MAXIS_case_number`
- [x] Update all instances of `footer_month` to read `MAXIS_footer_month`
- [x] Update all instances of `footer_year` to read `MAXIS_footer_year`
- [x] Merge stats block above the FuncLib (removing #2042 because it's on this issue now)
- [x] Replace `worker_county_code_determination` with `get_county_code`

theVKC:
OK, updating this to reflect a lot more work that should be done.

I'd like to do the following:
- [ ] Update FuncLib block to remove the term "Alpha Users" and shorten the MsgBox (IT folks don't need to be involved, errors seldom come up and I am not hearing of HTTP errors)
- [ ] Update all instances of `case_number` to read `MAXIS_case_number`
- [ ] Update all instances of `footer_month` to read `MAXIS_footer_month`
- [ ] Update all instances of `footer_year` to read `MAXIS_footer_year`
- [ ] Update all instances of `PMI` to read `MAXIS_PMI`
- [ ] Update all instances of `HH_memb` to read `MAXIS_HH_memb`

By standardizing these variables across all scripts, it allows them to take on global multi-script usage... something important for Projects Bowie and Gorbachev. I have scriptwriter training scheduled for March 29-31 and would like this done by then. So, instead of a Monday hackathon, I might do this prior to a hackathon and update the master branch a good week-or-so before a release (for testing purposes).

theVKC:
I am going to move this to May to coincide with our work in Project Bowie.

theVKC:
Removing #2042, it is now a duplicate of this issue now that I've expanded the scope of this one.

PR #2204, opened 2016-05-10
#1938: case_number is now MAXIS_case_number

theVKC:


theVKC:
I have backed up master. If there are any major issues, we can restore from the backup. Here we go.

PR #2211, opened 2016-05-11
RESOLVES #2123 adding cutoff date calculating

### LINKED ISSUE #2123: enhancement: Postponed wreg verif WCOM
It has come to my attention that TIKLing to close for the verif due date may lead to cases staying open longer than needed. Per TE02.05.70.01 for cases where client's have used all their abawd months we need to close them on the appropriate date (to me this originally meant the 10 day return date) which appears to be close the case for the 1st of the 2nd month if application date is before 15, and 1st of 3rd month if application is post 15th.

I see a few solutions.

Creating a list of cutoff dates in FUNCLIB that we maintain
Creating something to calculate cutoff dates (could be tricky)
Changing the TIKL to occur based on application date (so if pre 15 then current month/day TIKL if post then next month on the 1st):
"Close case by cutoff this month for postponed verifications not returned(Allow to cutoff for client to turn them in)"

Thoughts?

CDPotter:
BLIP: Script will now TILK for next cutoff date to close case. Since this script is designed for the postponed wreg verification workaround we are supposed to close the case in accordance with expedited processing. This means sometimes closing cases before a ten day period to turn in verifications (to meet cutoff requirements).

### DIFF
--- a/Script Files/NOTICES/NOTICES - POSTPONED WREG VERIFS.vbs
+++ b/Script Files/NOTICES/NOTICES - POSTPONED WREG VERIFS.vbs
@@ -66,7 +66,7 @@ BeginDialog case_number_dlg, 0, 0, 196, 85, "Postponed WREG WCOM"
 EndDialog
 
 BeginDialog WCOM_dlg, 0, 0, 196, 115, "Postponed WREG WCOM"
-  EditBox 55, 10, 60, 15, verif_due_date
+  EditBox 75, 10, 60, 15, ten_day_cutoff
   EditBox 55, 30, 60, 15, closure_date
   EditBox 70, 50, 115, 15, verifications_needed
   EditBox 75, 70, 60, 15, worker_signature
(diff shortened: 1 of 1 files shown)


theVKC:
@CDPotter, does your script contain footer_month or footer_year? I'm about to pull a fix which makes those MAXIS_footer_month and MAXIS_footer_year.

theVKC:
@CDPotter, I checked it and we should be good. Thanks for your work!

theVKC:
@CDPotter, do you have a policy citation for the cutoff thing?

CDPotter:
Page 5 of TE02.05.70.01

"TIKL to close the case for the appropriate date if the postponed verifications are not received. 

If the application date is on or prior to the 15th issue benefits for 1 month - close case 1st of the 2nd month. If the application date is after the 15th issue benefits for the 1st and 2nd months - close case the 1st of the 3rd month. "

To me this read if client applies 4/5/16 case must be set to close for 05/1/16 or 4/17/15 appl date must be closed for 06/1/16. (similar to current postponed verification processes but manual.)

The only thing I think it's a little vague on is allowing 10 days to get verif, but since it's a work around AND the client gets the 10 day notice (if closed on cutoff) then we are hitting that requirement?

theVKC:
I think we are. Thanks for the citation!

PR #2259, opened 2016-05-25
Resolved #2258 (partial): Fixed content check/PND1

### LINKED ISSUE #2258: Bulk Rept Lists (PND1, REVS, REVW) content error
The Rept Lists each check to see if a user "has_content" prior to carrying info to Excel. There seems to be little consistency in which content is checked for between these, but all seem okay except the above three. These three use the 2nd coordinate of the case number field and assume no case exists if it is blank. It misses cases of 5 or fewer numbers, and in doing so, misses entire caseloads if the first case alphabetically happens to have 5 or fewer numbers.

Proposing to switch this language to look for all 8 digits. I cannot publicly admit to doing this while my county has me on a script freeze... but if it were to accidentally be assigned to me, who knows what might happen?

Farleight:
REPT PND1 List has_content_check has been updated to look for an entire case number rather than one digit, eliminating missed cases on the bulk list.

### DIFF
--- a/Script Files/BULK/BULK - REPT-PND1 LIST.vbs
+++ b/Script Files/BULK/BULK - REPT-PND1 LIST.vbs
@@ -120,8 +120,8 @@ For each worker in worker_array
 	transmit
 
 	'Skips workers with no info
-	EMReadScreen has_content_check, 1, 7, 5
-	If has_content_check <> " " then
+	EMReadScreen has_content_check, 8, 7, 3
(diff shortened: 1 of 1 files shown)


theVKC:
@Farleight, why is this partial?

Farleight:
Because there are three scripts that have the same error. I'm working on the second now.

theVKC:
Oh, now I see!!

Farleight:
That should be all three now.

PR #2275, opened 2016-06-03
Resolves #2112 -- Creating MAGI/Non-MAGI Report

### LINKED ISSUE #2112: New Bulk Script - MAXIS HC Non-MAGI Persons/Cases
With the migration of MAXIS cases into METS starting with July 2017 HC ERs, we are looking at moving some cases around.  I would love a script that could be run on X1 numbers that would go through a caseload, check if the person is active on MA, coded as AA, AX, CB, CK or CX and check that the person has no spenddown.  If so, it would push that info to a bulk Excel spreadsheet so that we can determine how many persons/cases will be migrating and move cases accordingly. It would be nice if the script would also include their next HC ER date from the STAT/REVW panel.

RobertFewins-Kalb:
BLIP: This new script checks the inputted MAXIS caseloads for MAGI/Non-MAGI Health Care cases. Users have the option of adding the caseload's supervisor to the output. Additionally, users have the option of adding the status of CASH and SNAP on the MAXIS case.

### DIFF
--- a/Script Files/BULK/BULK - MAGI NON MAGI REPORT.vbs
+++ b/Script Files/BULK/BULK - MAGI NON MAGI REPORT.vbs
@@ -1,0 +1,416 @@
+'Gathering stats
+name_of_script = "BULK - MAGI NON MAGI LIST GEN.vbs"
+start_time = timer
+stats_counter = 1
+stats_manualtime = 67
+STATS_denomination = "C"
+
+'LOADING FUNCTIONS LIBRARY FROM GITHUB REPOSITORY===========================================================================
+IF IsEmpty(FuncLib_URL) = TRUE THEN	'Shouldn't load FuncLib if it already loaded once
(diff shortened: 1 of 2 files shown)


Script Files/BULK/BULK - MAGI NON MAGI REPORT.vbs
```diff
@@ -0,0 +1,416 @@
+'Gathering stats
+name_of_script = "BULK - MAGI NON MAGI LIST GEN.vbs"
```
theVKC: This is a different file name than the actual file submitted. Can you correct that for me, @RobertFewins-Kalb?

theVKC:
Boom! Thank you @RobertFewins-Kalb!

Script Files/BULK/BULK - MAGI NON MAGI REPORT.vbs
RobertFewins-Kalb: Twas a rookie mistake

PR #2282, opened 2016-06-07
Resolves #2246 - Bulk Case Transfer Bug

### LINKED ISSUE #2246: Bulk Script- Case Tranfer
I can get it to run until I try to put the new workers to transfer to in.
It errors out and says "Servicing Worker is Missing or Invalid" on the spec/xfer screen.
The worker number is showing correctly on the screen though.  
I've tried with x171xxx as Script states several times with several different worker numbers, also tried with just the 6 digits and last 3 digits to see if that made a difference and it won't work any of the ways I've tried.

Thanks,
Lisa

C-Love:
Blip - 'Fixes the script so that the cases will transfer without a MAXIS
error.'


theVKC:
@C-Love, would we consider this critical? Should this be resolved in release as well?

PR #2335, opened 2016-06-15
RESOLVES #2263 New script UTILITIES - COPY CASE/NOTE ELSEWHERE

### LINKED ISSUE #2263: Claims Script
A question was asked regarding claim notes.  Often times Eligibility Workers will just copy and paste the case notes for the claim into the claims note section.  Staff are wondering if they could get a script to put the same note into two different areas, CASE notes and CLAIM Notes.  Is that an option or does it already exist?

CDPotter:
BLIP: this script will allow users to copy a case note into claim notes or into a MEMO. NOTE: MEMO has a smaller space for writing so not all case notes will fit.

### DIFF
--- a/Script Files/UTILITIES/UTILITIES - COPY CASE NOTE TO.vbs
+++ b/Script Files/UTILITIES/UTILITIES - COPY CASE NOTE TO.vbs
@@ -1,0 +1,285 @@
+'Required for statistical purposes==========================================================================================
+name_of_script = "UTILITIES - COPY CASE NOTE TO.vbs"
+start_time = timer
+STATS_counter = 1               'sets the stats counter at one
+STATS_manualtime = 15           'manual run time in seconds
+STATS_denomination = "C"        'C is for each case
+'END OF stats block=========================================================================================================
+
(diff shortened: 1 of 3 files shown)


Script Files/UTILITIES/UTILITIES - COPY CASE NOTE TO.vbs
```diff
@@ -0,0 +1,285 @@
+'Required for statistical purposes==========================================================================================
+name_of_script = "UTILITIES - COPY CASE NOTE TO.vbs"
```
theVKC: This is a bit of an unusual file name. Perhaps "UTILITIES - COPY CASE NOTE ELSEWHERE.vbs" would be better? Ending a sentence with a "to" just feels wrong.

theVKC:
See note @CDPotter. Also, does it yell at you if you run out of room on a SPEC/MEMO? If not, it should.

CDPotter:
Name is fixed and it goes count up characters in the case note read and compares it to the max amount of characters a memo can accept.

PR #2390, opened 2016-07-01
RESOLVES #2389 correcting variable issue

### LINKED ISSUE #2389: Critical bug BULK INAC SCRUBBER unable to transfer cases
We put the wrong variable in when transferring cases causing the MAXIS_case_number variable to never change during a for/next. Fix incoming.

CDPotter:
BLIP: Script will now correctly loop through case number rather than getting stuck on last case number case/noted on.


CDPotter:
Side note: This has been tested on live case loads here at Anoka. This slipped through our fingers because DEV mode doesn't pause on the spec/xfer screen so we never saw that it was writing the same case number over and over.

theVKC:
@CDPotter, can you pull this into RELEASE as well?

Also, can you make a separate issue to ensure that dev mode pauses on this screen? Seems logical.

PR #2397, opened 2016-07-05
Resolves #2396 -- correcting age calculation bug

### LINKED ISSUE #2396: bug with Krabappel age calculation
I take the blame for this -- the way the training case generator is creating client's date of birth is incorrect. It is making clients a year too old. I have a fix ready.

RobertFewins-Kalb:
BLIP: Correcting a bug that was causing training clients to be created one year older than requested.


theVKC:
@RobertFewins-Kalb, see my comment on #2396. @C-Love, if you have more details, I'd love to hear them. :)

PR #2427, opened 2016-07-14
Resolves # 2371 - New Script FSS Status Change

C-Love:
Blip - 'New script that will update STAT based on information from a Status Update to categorize family cash cases as FSS. Case notes and TIKLs based on the information provided.'


theVKC:
@C-Love, I think you need to fetch upstream with the list-of-scripts. I'm getting a merge conflict.

PR #2455, opened 2016-08-11
Resolved #2454: Banked months FIATer ABAWD tracking record confirmation.

### LINKED ISSUE #2454: ABAWD banked months fiater bug fix: reading the initial banked month
Currently unless the month selected in the initial dialog is not coded as a "M", then the script errors out and says that the initial banked month is not coded on the ABAWD tracking record. This is an issue if there is an update to the case, the FIATer needs to be used, and the ABAWD tracking record is coded as an "x"

Assign to me for the August release. Pulling in now.

IlseFerris:
BLIP: Updated the script to account for cases that have the initial month coded as an "X" on the ABAWD tracking record.

### DIFF
--- a/Script Files/ACTIONS/ACTIONS - ABAWD BANKED MONTHS FIATER.vbs
+++ b/Script Files/ACTIONS/ACTIONS - ABAWD BANKED MONTHS FIATER.vbs
@@ -219,8 +219,14 @@ For each member in ABAWD_member_array 'This loop will check that WREG is coded c
   IF WREG_months < 3 THEN err_msg = err_msg & vbCr & "Member " & member & " has not used 3 ABAWD months in the past 3 years."
 	row = current_yr_row
 	col = current_month_col
-	EMSearch "M", row, col 'This looks to make sure there is an initial banked month coded on WREG on or after the initial month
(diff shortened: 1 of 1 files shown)


Script Files/ACTIONS/ACTIONS - ABAWD BANKED MONTHS FIATER.vbs
```diff
@@ -219,8 +219,14 @@ For each member in ABAWD_member_array 'This loop will check that WREG is coded c
   IF WREG_months < 3 THEN err_msg = err_msg & vbCr & "Member " & member & " has not used 3 ABAWD months in the past 3 years."
 	row = current_yr_row
 	col = current_month_col
-	EMSearch "M", row, col 'This looks to make sure there is an initial banked month coded on WREG on or after the initial month
-	IF row < current_yr_row OR row > current_yr_row + 1 THEN err_msg = err_msg & vbCr & "Member " & member & " does not have an initial banked month coded on WREG."
+	EMReadScreen tracking_record_coding, 1, row, col
```
CDPotter: Looks good but could we comment why we did this change and why we left the old code in commented out?

CDPotter:
Looks good but could you add some comments for it, and why we kept the old code?

PR #2515, opened 2016-09-12
Resolves #2500 New CSES Scrubber not able to note

### LINKED ISSUE #2500: DAIL - CSES Scrubber (the new one) not case noting.
In testing this morning, multiple instances have been found where the script can't open a case note, because it hasn't made it back to the dail.  It happens when it is left on the UNEA panel for worker review, or a worker does something in MAXIS while a dialog is up.   Probably needs some sort of check to make sure it is on the dail before opening the note.

courtrightd:
The script is now case noting correctly after reviewing UNEA panels.  Added logic to check if it made it to note, and fixed typos in case note function call.


CDPotter:
Thank you @courtrightd !

PR #2566, opened 2016-09-30
Resolves #2565 -- Fixing CASE/NOTE

### LINKED ISSUE #2565: BUG: REVW/MONT Closure case noting about inactive programs
Staff are reporting that BULK - REVW/MONT Closure is case noting about programs that are not up for review. An example is, for a case with no cash...

*Client has until   to turn in CAF/CSR and/or proofs for cash.

See case notes on the following MAXIS cases...
* 1018420
* 1468005
* 871235

I have a fix.

RoberT-1000:
BLIP: Adding logic to avoid case noting confusing lines about programs closing when those programs are not on the case.


CDPotter:
Looking ahead I'm going to say this one isn't critical. 10/16 release is due out 10/24 which will be one day after this script can be run.

PR #2568, opened 2016-09-30
Resolves #1142 -- Creating Inverse Krabappel

### LINKED ISSUE #1142: Krabappel: inverse Krabappel
Basically, creates the Excel spreadsheet from an existing case. Casey from Ramsey has offered to take this one on. Welcome to the Krabappel universe.

RoberT-1000:
BLIP: Adding the "Inverse Krabappel" script file. This script will grab case data from a training case and/or production cases, and will put the data into an Excel spreadsheet. For the script to run all the way through, you need to use a Krabappel template. The script is programmed to run with the data validation on the cells. INSTRUCTIONS TO COME!!!


CDPotter:
Could you update utilities main menu with a button?

PR #2584, opened 2016-10-04
Resolves #2583 -- Adding Wage Match Scrubber to DAIL Scrubber

### LINKED ISSUE #2583: New Script -- Wage Match (wink wink) DAIL scrubber
Anoka has been running a script that brings "wage match" and JOBS information for the quarter in question and the budget going forward to the financial worker similar to the BNDX scrubber. It's a good'un and I have been meaning to share for a while. The output looks like this...

image

And then the user gets to make their own decision.

RoberT-1000:
BLIP: This new script helps workers process WAGE DAILs. The script grabs quarterly earnings information from the match as well as earned income information for the case and presents that information in an organized dialog to the financial worker. The script does not update or case note anything, but it does save the financial worker time and improves the accuracy of transcribing earned income information from the matches.

### DIFF
--- a/COMPLETE LIST OF SCRIPTS.vbs
+++ b/COMPLETE LIST OF SCRIPTS.vbs
@@ -831,6 +831,15 @@ script_array(script_num).subcategory            = array("")
 script_array(script_num).release_date           = #10/01/2000#
 
 
+script_num = script_num + 1						'Increment by one
+ReDim Preserve script_array(script_num)			'Resets the array to add one more element to it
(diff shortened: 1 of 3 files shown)


CDPotter:
Name of script doesn't match the stats bit with name_of_script Also please add to DAIL scrubber script, and I think we're good. I've used this one and it's super handy!

RoberT-1000:
Will do...I'm updating the DAIL Scrubber file and then I'll go back through and update the script file.

RoberT-1000:
Finished

PR #2667, opened 2016-10-17
Resolves #2666 New CS Scrubber stuck in loop on MFIP / SNAP cases

### LINKED ISSUE #2666: New CSES scrubber stuck in loop on MFIP / SNAP combined cases
The script cannot finish it's case note and becomes stuck in an infinite loop on certain MFIP / SNAP households, because it attempts to read SNAP budget from MFIP sheet.  I should have a fix for this coming shortly.

courtrightd:
Blip: This resolves an error that was causing the script to be stuck in a loop while trying to note certain SNAP / MFIP combined household cases.


courtrightd:
This has been committed to a branch with another pull request.  Closing this.

PR #2685, opened 2016-10-24
RESOLVES #2682 correcting onesource link

### LINKED ISSUE #2682: Onesource Link
There is a new link for Onesource and the scripts currently have the old one. Is this on an agenda somewhere to be fixed and I have missed it?

CDPotter:
People will need to update their power pads to get this change. Luckily
the link currently points to a page where the new link is available so
it isn't super urgent.


CDPotter:
Please note that anyone who wants this new power pad will need to update their power pad file locally.

PR #2699, opened 2016-11-01
Resolves #2578 - Options for Notes on Income

### LINKED ISSUE #2578: Enhancement NOTES Explanation of income budgeted: self employment budget method noting
One of the issues ME reviewers are seeing is workers not documenting that the client agreed to the self employment budget method used. Since the CAF and other similar scripts are rather bloated I'd hate to add a checkbox for something like this. My thought was to add a checkbox to the explanation of income budgeted script. 

Maybe something like this:

capture

That case notes: "\* Client agreed to self employment budgeting method" if checked. 

Thoughts?/Anyone want it?

C-Love:
Blip - 'In order to provide better detail about income and how it is budgeted, there is now a button on the 'Notes on Income' area of the CAF, CSR, and Explanation of Income Budgeted' script that will allow workers to select from some predetermined options.'

### DIFF
--- a/notes/caf.vbs
+++ b/notes/caf.vbs
@@ -127,7 +127,7 @@ EndDialog
 BeginDialog CAF_dialog_02, 0, 0, 451, 305, "CAF dialog part 2"
   EditBox 70, 55, 370, 15, earned_income
   EditBox 80, 75, 360, 15, unearned_income
-  EditBox 110, 95, 330, 15, notes_on_income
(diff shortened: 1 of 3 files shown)


CDPotter:
Gonna run it by some people in ME first but I like it!

IlseFerris:
Is there a policy issue for the options that users can select? I thought we talked about the options already?

CDPotter:
Not that I'm aware of but I want to make sure all verbiage jives.

PR #2779, opened 2016-12-06
Fixing appointment time bug

CDPotter:
Appointment time variable was not being written if case wasn't phone interview.


CDPotter:
@RoberT-1000 let me know when this is working and I'll throw in release.

PR #2789, opened 2016-12-12
#649 Notes - appeals: case note script for appeal process

IlseFerris:
Added for Charles Clark who will be in MNsure training for the unforeseeable future
Will add his instructions to SIR this morning as well


CDPotter:
This one hast 5 dialogs in it. I'm wondering if we should rework it to be dynamic and using the dialog dialog (defining the dialogs just before the call) approach.

IlseFerris:
I don't know if that's totally necessary right now, but I can update this after the holidays if that's what you want to see.

CDPotter:
Since it's pushing REAL close to the limit (5+menu) I'd probably feel better if we went about it that way. I'll add a comment to the issue. Other than that it was looking like a good script! Thanks @IlseFerris and @senhorc !

PR #2810, opened 2016-12-28
resolves #2806 MFIP Sanction / DWP disqual policy issues

### LINKED ISSUE #2806: MFIP/DWP Disqual Notes Script - Policy Question
I have a worker who found the "last day to cure sanction" date for DWP incorrect.  Clients have until the last day of the month to cure the disqual for DWP.  Worker states “The policy referenced is for MFIP only. DWP does not have the same rules. The client always has until the last of the month to cure a DWP disqual and may cure it into the next month if good cause exists. Sanctions do not exist with DWP so the consecutive sanction policy is not applicable.”

Does the script need to be split into 2 scripts???

courtrightd:
Added correct handling of DWP disqual and improved noting.  New Sanction status for DWP disqual that makes sanction percent, number of sanctions, and pre/post 60 not mandatory fields when selected.   Also removes calculation of last day to cure sanction for DWP cases, as there is technically no last day to cure for DWP.

### DIFF
--- a/Script Files/NOTES/NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs
+++ b/Script Files/NOTES/NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs
@@ -38,11 +38,24 @@ IF IsEmpty(FuncLib_URL) = TRUE THEN	'Shouldn't load FuncLib if it already loaded
 END IF
 'END FUNCTIONS LIBRARY BLOCK================================================================================================
 
+'CHANGELOG BLOCK ===========================================================================================================
+'Starts by defining a changelog array
+changelog = array()
+
+'INSERT ACTUAL CHANGES HERE, WITH PARAMETERS DATE, DESCRIPTION, AND SCRIPTWRITER. **ENSURE THE MOST RECENT CHANGE GOES ON TOP!!**
+'Example: call changelog_update("01/01/2000", "The script has been updated to fix a typo on the initial dialog.", "Jane Public, Oak County")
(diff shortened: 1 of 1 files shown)


Script Files/NOTES/NOTES - MFIP SANCTION AND DWP DISQUALIFICATION.vbs
```diff
@@ -217,21 +232,26 @@ If action_type = "Apply sanction/disq."	then
...
+	IF sanction_status_droplist <> "DWP disqual" THEN
 	IF consecutive_sanction_months = checked then
 		Resolution_date = DateAdd("d", -1, Date_Sanction)
 	ELSEIf (sanction_type_droplist = "CS") then
 		Resolution_date = DateAdd("d", -1, Date_Sanction)
 	ELSE
 		Resolution_date = DateAdd("d", -10, Date_Sanction)
 	End If
-	
+	END IF
```
CDPotter: Just to make sure, we if it's a DWP disqualify we don't want to define/case note resolution date?

courtrightd: Correct - for DWP, the sanction can be cured at anytime during the 4 month DWP period, and benefits prorated from the date they are compliant, therefore there is no "resolution date".

CDPotter: Great! Thanks for the quick work on this one!

PR #2812, opened 2016-12-29
changed it was in the wrong directory

### LINKED ISSUE #2806: MFIP/DWP Disqual Notes Script - Policy Question

courtrightd:
Original pull was entered in wrong directory, please pull this in here


CDPotter:
I keep forgetting to check that... Also please don't add the resolves bit anymore as it will autoclose the issue and we want to leave them open now while testing. Thanks for catching this!

Issue #2846, opened 2017-01-17
Bug >> NOTES - CSR Case noting Verifs Needed twice


RoberT-1000:
The case note for the CSR includes an additional line where it is noting Verifs Needed twice. In the image below, the script is case noting the first instance of `verifs_needed` when it is not blank. At the bottom, it is case noting `verifs_needed` again.

I'll do it

![image](https://cloud.githubusercontent.com/assets/22157633/22037466/d9586550-dcbc-11e6-910e-12b958b9c08c.png)

PR #2847, opened 2017-01-17
RE #2846 Updating Case Noting in CSR

RoberT-1000:
BLIP: The script has been updated to clean up the case noting. It was case noting Verifs Needed twice.


CDPotter:
This is live in MASTER could I get confirmation that it is working? @MN-Script-Team/maxis-scriptwriters

IlseFerris:
Confirmed, working in master

Issue #2714, opened 2016-11-14
CSES DAIL scrubber enhancement: Add option to update for HC only cases


IlseFerris:
At hackathon Stearns, Ramsey, Hennepin and DHS discussed adding an option to users to update Child Support income for HC only cases. The justification for this: for MFIP cases that are temporarily closed due to incomplete HRF, users would need to process the income updates manually, unless the DAIL message was held until a later date. This effects case banking agencies like Stearns and Hennepin more than other agencies, but this is a solid option for all users. A vbYesNo can be added to give the user the option to update this income information or not.

courtrightd:
I will take this one too.

CDPotter:
@courtrightd please take a chance to test the updates Casey made to the scrubber last week while working on this. #2802

CDPotter:
@MN-Script-Team/maxis-scriptwriters Can I get confirmation this is working? Just need one agency for this smaller fix.

PR #2854, opened 2017-01-19
CSES scrubber update for HC only cases #2714

courtrightd:
Adds msgbox with yes/no to ask the user whether they want to update on cases where SNAP / MFIP are not open.


CDPotter:
Can you add a change log?

courtrightd:
Oops, done now.

Issue #2630, opened 2016-10-11
Asset assessment enhancement: add handling for more than 4 asset panels in ASET


IlseFerris:
Currently the script can handle up to 4 pages in the ASET function. There are 8 panels, so the logic needs to be reworked to account for reading and case noting all these panels.

CDPotter:
A few last minute fixes were added, can I get confirmation that this is working in MASTER @MN-Script-Team/maxis-scriptwriters (hopefully last test!)

Issue #2629, opened 2016-10-11
Asset assessment enhancement: add option for AA received 


IlseFerris:
AA are now required for disability waiver clients. This means that more and more AA will be coming into counties. Having an option to case note what was rec'd, what was needed, and some other basic information will be very helpful. 

I can take this for the November release unless someone has time before that to dedicate to it.

CDPotter:
Are you planning on re-naming it since it won't be exclusive to LTC?

PR #2859, opened 2017-01-20
#2629, #2630: AA updates

IlseFerris:


CDPotter:
@MN-Script-Team/maxis-scriptwriters could I get confirmation this is working in MAXIS? I have a fake case with ASET panels I could share as long we give it back to x127GM1 once done testing!

Changes: Added 'New AA received' option, and updated functionality of script to include enhanced password handling, and handling for all 8 pages of total marital assets.

PR #2881, opened 2017-02-15
Issue 2716 EVF Received Updated

MelissaFox-Stearns:
Added employer name to Tikl. 
(Initial pull request did not contain the update.)


C-Love:
@theVKC - this pull is just a verbiage change for the TIKL wording and should be a simple change. One note on the pull is that there is no changelog update.

This is definitely an enhancement and not a bug.

C-Love:
Confirmed working

Issue #2880, opened 2017-02-15
NOTES - MFIP SANCTION AND DWP DISQUALIFICATION - incorrect case note info.


StormageddonDLOA:
When selecting the checkbox that a SPEC/LETR was sent the case note reads that the client was notified via SPEC/WCOM.  Updating to fix the case noting issue.

PR #2882, opened 2017-02-15
issue #2880 MFIP SANCTION AND DWP DISQUALIFICATION - incorrect case n…

StormageddonDLOA:
…ote info

Updated the script to enter the case note correctly.


C-Love:
@theVKC - this code is correct except that it has been updated in the wrong file. This file us the ALL CAPS version that is no longer accessed. This would need to be redone on the correct case file. 

This is a bug that is creating incorrect case note information. If we think this is critical, I can update this work in a new pull request. I recommend closing this one.

theVKC:
#3011 pulled this in.

Issue #2944, opened 2017-06-29
Bug -- notes - lep sponsor income guidelines


RoberT-1000:
The script NOTES - LEP Sponsor Income includes income guidelines that have changed...

![image](https://user-images.githubusercontent.com/22157633/27707397-ced76a60-5cda-11e7-878c-ee60a835731c.png)

![image](https://user-images.githubusercontent.com/22157633/27707421-ebcfc7f2-5cda-11e7-8cfd-d673ec5576ca.png)

...and require updating.

MiKaylaH22:
We have had multiple requests on this script, if this can be updated would be okay to change some of the language in the script?

Farleight:
Which language would you like changed? If you're thinking of changing it from English to Japanese, sore wa ii desu ne.

theVKC:
Ok folks, I'm pulling this in tonight for testing, on behalf of MAXIS who does not currently have a scriptwriter. I'll be taking over the approval of any bug fixes in anticipation of MAXIS going to a new FuncLib for Win10. More to come on that later. Do not fear. :)

@MiKaylaH22, your idea is noble. If you want to change the language *to resolve a bug*, please let me know and feel free to submit a pull request. Keep in mind that a new version of this script is about to be pulled in, so please please please ensure you are working off of that version (you'll see it tagged below). If you haven't created a new issue for your recommendation for language update, please do so, as we will track this separately (this is a critical issue and I want it off the to-do list).

theVKC:
OK, so this issue is a bit old... but here's what I have based on info from @C-Love: please correct me if I'm wrong...

```vb.net
If sponsor_HH_size = 1 then income_limit = 1307
If sponsor_HH_size = 2 then income_limit = 1760
If sponsor_HH_size = 3 then income_limit = 2213
If sponsor_HH_size = 4 then income_limit = 2665
If sponsor_HH_size = 5 then income_limit = 3118
If sponsor_HH_size = 6 then income_limit = 3571
If sponsor_HH_size = 7 then income_limit = 4024
If sponsor_HH_size = 8 then income_limit = 4477
If sponsor_HH_size > 8 then income_limit = 4430 + (453 * (sponsor_HH_size - 8))
``` 

This is what is being pulled in from https://github.com/C-Love/DHS-MAXIS-Scripts/blob/664d07285d6b22eaceb9dfa62331a64b17858de2/notes/lep-sponsor-income.vbs. :)

[...]

PR #2987, opened 2017-11-13
#2944: sponsor income limit update

theVKC:
Note that this originally was done by @c-love but because GitHub was acting weird I had to redo it.

Testers, are you ready? @C-Love? @IlseFerris?


IlseFerris:
Line 113 needs to be updated. Multiplier is 4477, not 4430

PR #3020, opened 2018-02-27
resolves #2966

### LINKED ISSUE #2966: QUESTION: Scripts updated to work with the modified new hire messages?
I'm not sure how or who to even submit this question to since this repository probably isn't being actively monitored. One of my workers has a question so I thought I'd put it on here just in case someone sees it :) She is wondering if scripts will be updated to work with the modified new hire messages? If someone sees this and is able to respond, great. If not, it's a-ok.

StormageddonDLOA:
Accounts for the updates to the NDNH Dail message to allow the script to run.

Adds functionality to add a new JOBS panel in the future month if it wouldn't carry over on its own.

Adds functionality to resolve the HIRE match through IVES only on the jobs that are known, where there won't be any future savings, by going to IVES, allowing the worker 5 seconds to select the correct job listed (in case there is more than one) then codes it as known and removes it.

### DIFF
--- a/Script Files/DAIL/DAIL - NEW HIRE NDNH.vbs
+++ b/Script Files/DAIL/DAIL - NEW HIRE NDNH.vbs
@@ -82,22 +82,33 @@ If len(current_day) = 1 then current_day = "0" & current_day
 current_year = datepart("yyyy", date)
 current_year = current_year - 2000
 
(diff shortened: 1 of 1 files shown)


theVKC:
Because @C-Love noted the error in the file name (something which will be repaired in Python), we will be closing this pull request. However, @StormageddonDLOA, your actual work looked great. Not your fault the dual files are confusing.

PR #3022, opened 2018-02-27
Resolved #2897

### LINKED ISSUE #2897: NOTES - EMERGENCY bug for x127/x162 users: header and crisis info duplicated
If a user in Hennepin or Ramsey county uses the EGA screening functionality, then continues on to use the NOTES - EMERGENCY script, the crisis info from the screening is carried over which causes the header of the case note and the 'crisis' field to duplicate. 

Example case note header: ***Emergency app: homelessnesshomelessness***  

The crisis variable needs to be blanked out prior to going into the NOTES - EMERGENCY script around line 396.

MelissaFox-Stearns:
Blanked out crisis variable for Hennepin and Ramsey users to heading is not duplicated in case note.

### DIFF
--- a/notes/emergency.vbs
+++ b/notes/emergency.vbs
@@ -394,7 +394,8 @@ If EGA_screening_check = 1 then
 	END IF
 END IF
 'End of EMER screening code----------------------------------------------------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


theVKC:
Closing per @IlseFerris. May be the wrong file.

Issue #2971, opened 2017-09-28
Bugs on ACTIONS - FSS Status Change


C-Love:
Script has a few errors:

- Second dialog length is miscalculated - need to extend so the OK and Cancel buttons are visible (buttons to continue the script run are always nice)

- Sometimes the script throws a date error 

- No passworded out protection at the beginning of script run

I have fixes for these prepared.

C-Love:
We are debating about if this is worth our time to update prior to conversion. @MN-Script-Team/maxis-scriptwriters  - does anyone use this script?Is this a problem for workers?

theVKC:
Please let me know ASAP if this is tested! Thank you!

C-Love:
Tested in Master - working and looks good.

PR #3028, opened 2018-03-01
Issue #2971 - Bugs on FSS Status Change

C-Love:
Fixing bugs - dialog length, date format issue, and a check for background.


C-Love:
@theVKC  - yes it appears that was from testing - honestly I ran it through to test it but I didn't write this code recently.

